PR #67, opened 2026-04-03
Allow installing packages when same name exists on another channel

### LINKED ISSUE #53: BUG: if package is installed (eg locally or on a different channel) then does not permit installing from eg main channel...
get the error:

Package "local/local/hello_mip" is already installed. Uninstall first to reinstall.

magland:
Fixes #53

## Summary
- Explicitly requested packages are now installed even if the same bare name exists on another channel (e.g., `local/local/hello_mip` no longer blocks installing `mip-org/main/hello_mip`)
- Dependencies are still satisfied by any channel (bare name match), preserving existing behavior
- Adds support for FQN dependencies in `mip.yaml` — a bare name dep (`foo`) is satisfied by any channel, while an FQN dep (`mip-org/main/foo`) requires that exact installation

## Changed files
- `+mip/install.m` — skip cross-channel blocking for explicitly requested packages
- `+mip/+utils/install_local.m` — handle FQN deps when checking if dependencies are installed
- `+mip/+dependency/build_dependency_graph.m` — extract bare name from FQN deps for index lookup
- `+mip/+dependency/topological_sort.m` — extract bare name from FQN deps for sorting

### DIFF
--- a/+mip/install.m
+++ b/+mip/install.m
@@ -49,16 +49,12 @@ function install(varargin)
     end
 
     % Check if any argument is a local directory with mip.yaml
-    % Only treat as local path if it looks like one (contains separator,
-    % starts with '.' or '~'), not for bare package names or FQNs
     for i = 1:length(args)
         pkg = args{i};
-        looksLikePath = startsWith(pkg, '.') || startsWith(pkg, '~') || ...
-                        startsWith(pkg, '/') || startsWith(pkg, '\') || ...
-                        contains(pkg, [filesep filesep]);
-        if looksLikePath && exist(pkg, 'dir')
+        % Resolve '.' and relative paths
+        if isfolder(pkg)
             mipYamlPath = fullfile(pkg, 'mip.yaml');
-            if exist(mipYamlPath, 'file')
+            if isfile(mipYamlPath)
                 mip.utils.install_local(pkg, editable);
                 return;
             else


magland:
## Latest changes: full cross-channel install support

### Require org/channel format everywhere

`parse_channel_spec` no longer accepts bare names like `'core'` or `'dev'`. All channel references must be fully qualified as `'org/channel'` (e.g. `'mip-org/core'`). All internal defaults updated from `'core'` to `'mip-org/core'`. This removes an implicit assumption and makes the channel identity explicit throughout the codebase.

Files changed: `parse_channel_spec.m`, `install.m`, `update.m`, `index.m`, `avail.m`, `info.m`, `resolve_package_name.m`, `fetch_index.m`

### Test channels created

Two new GitHub repos with inline test packages:

**mip-org/mip-test-channel1** — `alpha` (1.0.0, 2.0.0), `gamma` (FQN dep on `mip-org/test-channel1/alpha`), `shared` (1.0.0), `core_dep` (bare-name dep on `chebfun`)

**mip-org/mip-test-channel2** — `beta` (1.0.0), `cross` (FQN dep on `mip-org/test-channel1/alpha`), `shared` (1.0.0, 2.0.0)

### Three install fixes

**1. FQN install fetches the right channel**

Previously `mip.install('mip-org/test-channel1/alpha')` would only fetch the default channel (core) index and fail because the package isn't there. Now `installFromRepository` collects all unique channels referenced by the resolved FQNs and fetches each one's index upfront.

**2. Cross-channel FQN dependencies auto-fetched**

`build_dependency_graph` no longer silently skips dependencies not in the `packageInfoMap`. Instead it errors with `mip:packageNotFound`. `install.m` wraps the dep graph building in a retry loop — on that error, it parses the missing FQN, fetches its channel's index, merges into the map, and retries. If the channel was already fetched, the error is re-thrown (package truly doesn't exist).

This means `mip.install('--channel', 'mip-org/test-channel2', 'cross')` automatically fetches `mip-org/test-channel1` to install the `alpha` dependency.

**3. Local directory detection no longer triggers on bare names**

`exist('shared', 'dir')` was matching MATLAB toolbox directories, causing `mip.install('--channel', '...', 'shared')` to error with `noMipYaml`. The local directory check now only triggers for path-like arguments (starting with `.`, `/`, `~`, or `\`).

### Integration test file

`tests/TestChannelInstall.m` — 13 test cases exercising:
- Basic install from non-core channel (bare name + `--channel`, FQN, specific version)
- Same-channel FQN dependency resolution
- Cross-channel FQN dependency auto-fetch
- Cross-channel dep when pre-installed
- Same-named packages installed side-by-side from different channels
- Bare-name dependency resolving to `mip-org/core`
- Load from non-core channel (basic, function accessible, with deps)
- Load disambiguation by FQN when same name on multiple channels
- Idempotent re-install

Conditionally included in `run_tests.m` — skipped when `MIP_SKIP_REMOTE` is set.

### Dependency resolution principle

- **Bare-name dependency** = `mip-org/core/<name>` (always)
- **Same-channel dependency** (non-core) must use FQN
- **Cross-channel dependency** must use FQN
- Only on `mip-org/core` do bare names "just work" since bare = core

### Testing

All 13 channel tests pass. All 137 existing tests pass (124 original + 13 new parsing tests).

PR #84, opened 2026-04-03
Fix websave timeouts and skip @/+ dirs in recursive addpath

magland:
## Summary

- Set 60s timeout on all `websave` calls (`fetch_index`, `download_mhl`, `avail`, `info`). The default 5s timeout causes failures on slower connections when downloading packages or fetching indices from GitHub.
- Skip `@ClassName` and `+namespace` directories in `compute_addpaths` recursive path discovery, matching MATLAB's `genpath` behavior. These directories are auto-discovered by MATLAB from their parent and produce warnings when added to the path explicitly.

cc @danfortunato

### DIFF
--- a/+mip/load.m
+++ b/+mip/load.m
@@ -6,6 +6,7 @@ function load(varargin)
 %   mip.load('package1', 'package2', ...)
 %   mip.load('packageName', '--sticky')
 %   mip.load('packageName', '--install')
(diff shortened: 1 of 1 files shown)


magland:
@danfortunato I am merging this because I need it for preparing some pages, but feel free to look over and comment.

PR #87, opened 2026-04-06
Add mip test command

magland:
## Summary
- Adds `mip test <package>` command that loads the package and runs its `test_script` from `mip.yaml`
- Supports `test_script` at both top-level and per-build in `mip.yaml` (same as `compile_script`)
- Persists `test_script` in `mip.json` for editable installs
- Packages without a test script print a message and return gracefully


danfortunato:
LGTM

PR #108, opened 2026-04-08
Add test_script so 'mip test mip' runs the unit tests

magland:
## Summary
- Adds `test.m` at the package root that invokes `tests/run_tests.m` and errors on any failed test.
- Sets `test_script: test.m` in `mip.yaml` so `mip test mip` picks it up.

Tested by installing my dev mip as `local/local/mip` and then running `mip test local/local/mip`.

Closes #106

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -8,7 +8,11 @@ on:
 
 jobs:
   test:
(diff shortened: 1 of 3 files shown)


danfortunato:
LGTM

PR #110, opened 2026-04-08
Fully ignore --channel when all install args are FQN

magland:
## Summary

Resolves #105. When every `mip install` argument is a fully qualified name, `--channel` is now ignored *entirely* — no spurious index fetch, no misleading "Using channel" line. Mixed bare+FQN calls are unchanged. The silent-ignore behavior for FQN args was confirmed in the issue thread.

## Changes

- [+mip/install.m](+mip/install.m): track `hasBareName` while resolving args. Skip the primary-channel fetch and the "Using channel" print when no bare-name arg is present. `mip-org/core` is still fetched unconditionally so bare-name dependencies can resolve.
  - §3.1.1 / §3.1.2: documented the new index-fetching rule.
  - §14.18: rewrote from open question to documented decision; removed the incorrect "still affects the channel index fetch for bare-name dependencies" claim (bare-name deps always resolve to `mip-org/core` per §3.1.5).
- [tests/TestInstallRemote.m](tests/TestInstallRemote.m): three new tests covering FQN-with-`--channel`, FQN-only with bogus `--channel` (proves the spurious fetch is gone), and mixed bare+FQN.

## Test plan

- [x] `runtests('TestUtilsParsing')` — 26/26 pass
- [x] `runtests('TestInstallRemote')` — 16/16 pass (13 existing + 3 new)

### DIFF
--- a/+mip/install.m
+++ b/+mip/install.m
@@ -140,18 +140,19 @@ function installedFqns = installFromRepository(repoPackages, ~, channel)
 
     [defaultOrg, defaultChan] = mip.utils.parse_channel_spec(channel);
 
-    fprintf('Using channel: %s/%s\n', defaultOrg, defaultChan);
-
-    % Get current architecture
-    currentArch = mip.arch();
-    fprintf('Detected architecture: %s\n', currentArch);
-
-    % Resolve each package argument to org/channel/name (with optional version)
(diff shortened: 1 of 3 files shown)


danfortunato:
LGTM

PR #111, opened 2026-04-08
Document local-update reinstall semantics as intentional (#103)

magland:
## Summary

Resolves #103. The current \`mip update local/local/<pkg>\` behavior — always uninstall + reinstall from source, always re-run \`compile_script\` for editable installs — is intentional. Documents the resolved decisions in [docs/behavior-reference.md](docs/behavior-reference.md) and adds two tests that pin the editable-recompile behavior.

## Changes

  - §7.2: explicitly notes that for editable installs, \`compile_script\` runs again on every update and the original \`--no-compile\` flag is not preserved.
  - §14.16: rewritten from open question to documented decision. Cites #103 and includes the forward-looking note that a future \`mip update --all\` should not include local packages.
- [tests/TestUpdateLocal.m](tests/TestUpdateLocal.m): two new tests using the existing \`compile_script\` helper:
  - \`testUpdateLocalPackage_EditableRerunsCompileScript\` — delete the compile-marker, run \`mip update\`, verify the marker is recreated.
  - \`testUpdateLocalPackage_NoCompileFlagNotPreserved\` — \`mip install -e --no-compile\` produces no marker; \`mip update\` produces one, proving the flag isn't carried across.

## Test plan

- [x] \`runtests('TestUpdateLocal')\` — 12/12 pass (10 existing + 2 new)

### DIFF
--- a/docs/behavior-reference.md
+++ b/docs/behavior-reference.md
@@ -494,7 +494,7 @@ Using an FQN bypasses this check entirely.
 5. Reinstall from source (preserving editable/non-editable mode).
 6. Reload if it was previously loaded.
 
-Local updates **always** reinstall (no up-to-date check). Timestamps change on every update.
+Local updates **always** reinstall (no up-to-date check). Timestamps change on every update. For editable installs the `compile_script` runs again on every update; the `--no-compile` flag from the original install is **not** preserved. See [§14.16](#1416-mip-update-on-local-package-always-reinstalls).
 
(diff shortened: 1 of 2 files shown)


danfortunato:
LGTM. Do we want a `--no-compile` flag for `mip update`?

magland:
> LGTM. Do we want a `--no-compile` flag for `mip update`?

Sure. I'll open an issue for that.

PR #112, opened 2026-04-08
Replace installed version when mip install is given an explicit @version (#102)

magland:
## Summary

Resolves #102. \`mip install pkg@2.0\` no longer prints "already installed" and does nothing when a different version of pkg is already on disk. When the user passes an explicit \`@version\` and a different version is currently installed, the old copy is silently replaced.

## Behavior

- **Explicit \`@version\`, different installed**: silently uninstall the old version, install the requested one. If the package was loaded, unload before and reload after.
- **Explicit \`@version\`, same installed**: idempotent — no action.
- **No \`@version\`**: unchanged. Still prints "already installed" and skips.
- **Replacement guard**: only triggers when the version mip *would* install matches the requested one. Prevents accidentally ripping out a working install when the version constraint isn't reaching the FQN's channel (documented as a known limitation in §14.13).

## Changes

- [+mip/install.m](+mip/install.m): track \`requested_version\` per resolved package; before the install loop, check each user-requested package — if a different version is installed and mip would install the requested one, unload + uninstall it. After the install loop, reload anything that was unloaded.
  - §3.1.7 documents the new \`@version\` exception with a cross-link to §14.13.
  - §14.13 rewritten from open question to documented decision; calls out the FQN-version-constraint bug as a known limitation and notes lock files / version constraints are out of scope.
- [tests/TestInstallRemote.m](tests/TestInstallRemote.m): five new tests covering upgrade, downgrade, idempotency, no-\`@version\` no-op, and load-state preservation.

## Test plan

- [x] \`runtests('TestInstallRemote')\` — 18/18 pass (13 existing + 5 new)
- [x] \`runtests('TestUpdateRemote')\` — 5/5 pass (no regressions)

### DIFF
--- a/+mip/install.m
+++ b/+mip/install.m
@@ -148,13 +148,14 @@ function installedFqns = installFromRepository(repoPackages, ~, channel)
 
     % Resolve each package argument to org/channel/name (with optional version)
     % The --channel flag only applies to user-specified bare names, not dependencies
-    resolvedPackages = {};  % cell array of structs with .org, .channel, .name, .fqn
+    resolvedPackages = {};  % cell array of structs with .org, .channel, .name, .fqn, .requested_version
     requestedVersions = containers.Map('KeyType', 'char', 'ValueType', 'any');
     for i = 1:length(repoPackages)
         pkg = repoPackages{i};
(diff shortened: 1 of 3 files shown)


danfortunato:
LGTM

PR #113, opened 2026-04-08
Roll back orphaned dependencies on failed install (#100)

magland:
## Summary

Resolves #100. If a multi-package install fails midway, the dependencies that already succeeded are now rolled back instead of being left as orphans. The implementation extracts the existing prune logic from \`mip uninstall\` into a shared utility and calls it from the install failure path.

## Behavior

- **Successful install**: unchanged.
- **Failed install (any package in the topological install loop fails)**: \`mip install\` runs \`mip.utils.prune_unused_packages\` and re-raises the original error. Because user-requested packages haven't been added to \`directly_installed.txt\` yet, anything installed by this call but not needed by an already-directly-installed package is pruned as an orphan.
- **\`.mhl\` install with remote deps that fails**: same path -- prune any orphans, re-raise.
- **\`mip uninstall\`**: unchanged behavior. The local \`pruneUnusedPackages\` was extracted, not changed.

## Changes

- New: [+mip/+utils/prune_unused_packages.m](+mip/+utils/prune_unused_packages.m) -- the prune logic, lifted from \`uninstall.m\`. The broken-dependency check stayed in \`uninstall.m\` because it's uninstall-specific.
- [+mip/uninstall.m](+mip/uninstall.m): replace local \`pruneUnusedPackages\` (and its \`getAllDependencies\` helper) with a call to the new utility, followed by \`checkForBrokenDependencies\` as before.
- [+mip/install.m](+mip/install.m): wrap the install loop in \`try/catch\`; on failure, call \`mip.utils.prune_unused_packages\` and re-raise. Same treatment for \`installFromMhl\`. If the prune itself fails, emit a \`mip:rollbackFailed\` warning and still re-raise the original error.
  - §3.1.6 documents the new failure-path behavior.
  - §14.11 rewritten from open question to documented decision; notes the (mild) caveat that the prune sweep also removes any pre-existing orphans.
- [tests/TestUninstallPackage.m](tests/TestUninstallPackage.m): five new tests pinning the extracted utility:
  - removes orphans
  - preserves directly-installed packages
  - preserves transitive deps
  - mixed orphan + directly-installed state
  - never prunes \`mip-org/core/mip\`

I did not add an integration test that triggers a real failed-install rollback -- making the install loop fail at a controlled point would require a deliberately-broken test channel package or mocking, which is more effort than it's worth for this PR. The existing \`mip uninstall\` test suite plus the new direct unit tests cover the prune logic; the wiring in \`install.m\` is small enough to review by inspection.

## Test plan

- [x] \`runtests('TestUninstallPackage')\` -- 24/24 pass (19 existing + 5 new)
- [x] \`runtests('TestInstallLocal')\` -- 10/10 pass
- [x] \`runtests('TestInstallRemote')\` -- 13/13 pass

### DIFF
--- a/+mip/+utils/prune_unused_packages.m
+++ b/+mip/+utils/prune_unused_packages.m
@@ -1,0 +1,132 @@
+function prune_unused_packages()
+%PRUNE_UNUSED_PACKAGES   Remove installed packages that are no longer needed.
+%
+% A package is considered needed if it is in `directly_installed.txt` or
+% is a transitive dependency of any directly-installed package.
+%
+% `mip-org/core/mip` (the package manager itself) is never pruned.
+%
+% Used by:
+%   - `mip uninstall`: prune orphans after removing requested packages.
+%   - `mip install`: roll back successfully-installed dependencies when a
+%     later package in the same operation fails.
+
+    allInstalled = mip.utils.list_installed_packages();
+
+    if isempty(allInstalled)
+        return
+    end
+
+    directlyInstalled = mip.utils.get_directly_installed();
+
+    % Build set of all needed packages (directly installed + their dependencies)
+    neededPackages = {};
+    for i = 1:length(directlyInstalled)
+        directPkg = directlyInstalled{i};
+        if ismember(directPkg, allInstalled)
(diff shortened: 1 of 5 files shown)


danfortunato:
LGTM

PR #116, opened 2026-04-09
Make `mip install <bare>` always prefer channel over local dir

magland:
Closes #107.

## Summary
- `mip install` now categorizes each argument up front: starts with `~`/`./`/`/` → local path; bare name or `org/channel/pkg` FQN → channel; anything with 2 or 4+ slash-separated parts is rejected with `mip:install:invalidPackageSpec` and a hint about prefixing with `./`.
- A bare name like `chebfun` always tries the channel, even if a `chebfun/` directory happens to exist in the current folder. To force a local install, the user must write `./chebfun`.
- If the channel install fails and one of the requested names also exists as a relative directory in cwd, the error message is augmented with a `./<name>` hint so the user knows how to install it as a local package instead.
- `--editable`/`-e` is now only valid when at least one local-path argument is present.

## Tests
- New offline tests in `TestInstallLocal.m` cover the parsing rules (length-2 / length-4+ rejection, `./mypkg` dispatch, `-e bare` rejection).
- New integration tests in `TestInstallRemote.m` verify that a colliding local `alpha/` directory does not shadow `mip-org/test-channel1/alpha`, and that a failing bare-name install surfaces the `./` hint when a same-named local directory exists.
- All 217 tests pass.

## Docs
- Added §3.0 *Argument Categorization* to `docs/behavior-reference.md` explaining the new dispatch rules.
- Updated §3.2 *Local Installation* to drop the old `isfolder()` wording.

### DIFF
--- a/+mip/install.m
+++ b/+mip/install.m
@@ -21,9 +21,14 @@ function install(varargin)
 %   --no-compile        Skip compilation (editable installs only)
 %
 % Local packages:
-%   If the argument is a directory path containing a mip.yaml file,
-%   the package is installed locally. In editable mode, changes to
-%   the source directory are reflected immediately without reinstalling.
+%   To install a local directory, the path must start with '~', '.', or
+%   '/' (e.g. './mypkg', '/abs/path/mypkg'). The directory must contain
+%   a mip.yaml file. In editable mode, changes to the source directory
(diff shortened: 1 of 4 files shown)


danfortunato:
LGTM

PR #119, opened 2026-04-09
Raise mip:loadError instead of marking failed load as loaded

magland:
## Summary

- If `load_package.m` errors, `mip.load` now raises `mip:loadError` (original error attached as cause) and the package is **not** added to `MIP_LOADED_PACKAGES` / `MIP_DIRECTLY_LOADED_PACKAGES`, so the user can fix the script and retry without first running `mip unload`.
- The error propagates through dependency recursion: a parent whose dependency failed to load is also not marked as loaded.
- The `mip.json` read is now wrapped in its own narrow try/catch so a parse failure still degrades to a warning, but recursive `loadSingle` calls are no longer silently swallowed by that catch.
- We do **not** attempt to roll back path mutations a partially-executed `load_package.m` may have made -- that is not generally feasible. The new docs call this out explicitly.

Behavior reference updated in §4.1, §4.6, §4.7, §13, and §14.17 (resolved). New tests in `TestLoadPackage` cover: error is thrown, package is not marked loaded, retry works, working directory is restored, and parent-not-loaded when a dependency fails.

Closes #104.

### DIFF
--- a/+mip/load.m
+++ b/+mip/load.m
@@ -121,38 +121,41 @@ function loadSingle(packageArg, installIfMissing, stickyPackage, channel, isDire
         return
     end
 
-    % Check for mip.json and process dependencies
+    % Check for mip.json and process dependencies. Only the parse step is
+    % wrapped in try/catch so that recursive dependency-load errors propagate
+    % instead of being silently downgraded to a warning.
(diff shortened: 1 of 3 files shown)


danfortunato:
LGTM

PR #120, opened 2026-04-09
Validate MIP_ROOT and document resolution rules

### LINKED ISSUE #101: `MIP_ROOT` env var: clarify empty-string and nonexistent-path handling
`MIP_ROOT` is used to override the default `~/.mip` root, but a couple of edge cases aren't well-defined.

## Current behavior

[root.m:9-12](+mip/root.m#L9-L12):
```
root = getenv('MIP_ROOT');
if ~isempty(root)
    return;
end
```

- `MIP_ROOT` **unset**: falls through to path-based detection.
- `MIP_ROOT=""` (set to empty string): `getenv` returns `''`, indistinguishable from unset → falls through to path-based detection. Same as unset.
- `MIP_ROOT=/nonexistent/path`: returned as-is, no validation. Subsequent calls to `get_packages_dir`, `list_installed_packages`, etc. operate on a non-existent directory and fail with whatever errors those calls produce.

## Questions to decide

1. Is "empty string == unset" the right behavior, or should empty string be an error?
2. Should `mip.root()` validate that `MIP_ROOT` (when set) points to an existing directory? If not, what's the user-facing message when it doesn't?
3. Should `mip.root()` create `MIP_ROOT/packages/` if missing, or refuse?

## Goal

Decide the rules for `MIP_ROOT` validation. Then update `mip/docs/behavior-reference.md` and tests to match.

magland:
Resolves #101.

## Summary
- `mip.root()` now errors with `mip:rootInvalid` when `MIP_ROOT` is set but the path doesn't exist, isn't a directory, or doesn't contain a `packages/` subdirectory. It does **not** auto-create `packages/`.
- Empty string remains indistinguishable from unset (per `getenv` semantics) and falls through to path-based detection.
- New `tests/TestMipRoot.m` covers all four cases plus a happy-path test.
- `docs/behavior-reference.md` gains a new §11.5 documenting the rules, marks §14.12 resolved, adds `mip:rootInvalid` to the error table, and replaces a stale `~/.mip` placeholder in the layout diagram.

## Test plan
- [x] `run_tests()` — 244 / 244 pass, including 5 new `TestMipRoot` tests


danfortunato:
LGTM

PR #132, opened 2026-04-10
Update packages in place instead of uninstall+install

### LINKED ISSUE #99: `mip update` doesn't check or update dependencies
`mip update <pkg>` rebuilds only the named package. Its dependencies are not revisited, even if newer versions are available in the channel index.

## Current behavior

[updateRemotePackage](+mip/update.m#L104-L181) does:

1. Fetch the channel index, find the latest variant for `<pkg>`.
2. If installed version + commit hash match latest, no-op (unless `--force`).
3. Otherwise: `rmdir(pkgDir, 's')` → `mip.install(fqn)`.

The `mip.install(fqn)` call does build a dependency graph, but for each dep it checks "is the directory already there?" ([install.m:259-268](+mip/install.m#L259-L268)) and if yes, prints "already installed" and skips. There is no version comparison against the index for those existing deps.

There is also no `mip update --all` or `mip update --recursive` command. The user must name each package they want to update.

## Questions to decide

1. Should `mip update <pkg>` check dependencies and offer/perform updates for ones with newer versions?
2. Should there be a `mip update --recursive <pkg>` (just this subgraph) and/or `mip update --all` (every directly installed package)?
3. If a dep is shared between multiple installed packages, should updating it be conditional on the others still being compatible?
4. Does this depend on first deciding version constraints (#95) and a lock file (#96)? Without constraints, "compatible" is undefined.

## Goal

Decide the recursive/transitive scope of `mip update`. Then update `mip/docs/behavior-reference.md` and tests to match.

magland:
## Summary

- Remote packages are now updated by replacing the directory on disk and downloading the new version, rather than going through `mip.uninstall` + `mip.install`
- Existing dependencies are left as-is (not reinstalled or updated)
- New dependencies required by the updated version are installed automatically
- Orphaned dependencies (no longer needed) are pruned
- Load state is still preserved across updates

Per discussion in #99 — this matches Homebrew-style behavior where `update X` only updates X, not its deps.

Closes #99

### DIFF
--- a/+mip/update.m
+++ b/+mip/update.m
@@ -17,17 +17,17 @@ function update(varargin)
 % considered to need an update. If a package does not need updating,
 % nothing happens for that package (unless --force is specified).
 %
-% For the remote packages that need updating, `mip update X Y Z` is
-% equivalent to `mip uninstall X Y Z` followed by `mip install X Y Z`.
-% This re-resolves each package's dependency graph against the current
-% channel index, so dependencies are effectively updated as well.
+% Remote packages are updated in place: the old directory is removed and
+% the new version is downloaded. Existing dependencies are not updated.
+% After the update, any new dependencies that the updated package requires
+% are installed, and any orphaned packages (old dependencies no longer
+% needed by any directly installed package) are pruned.
 %
 % Local packages are reinstalled directly from their source path without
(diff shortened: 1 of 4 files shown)


danfortunato:
LGTM

PR #135, opened 2026-04-10
feat: reapply cpu_level-aware variant selection and bundle support

danfortunato:
## Summary
- Reapplies the changes from #109 (reverted in #121)
- Adds `detect_cpu_level.m` for pure-MATLAB CPU SIMD level detection (Linux via `/proc/cpuinfo`, Windows via `kernel32`)
- `select_best_variant.m` picks the highest cpu_level variant that doesn't exceed host capability
- `bundle.m` gains `--cpu-level` flag; `create_mip_json.m` stores `cpu_level` field

## Changes from original PR
- `detect_cpu_level.m` placed in `+mip/+resolve/` (was `+mip/+utils/` which no longer exists after the refactor in #126)
- Updated namespace reference from `mip.utils.detect_cpu_level` to `mip.resolve.detect_cpu_level`

### DIFF
--- a/+mip/+build/create_mip_json.m
+++ b/+mip/+build/create_mip_json.m
@@ -53,6 +53,10 @@ end
 
 mipData.architecture = architecture;
 
+if isfield(opts, 'cpu_level') && ~isempty(opts.cpu_level)
+    mipData.cpu_level = opts.cpu_level;
+end
+
 if isfield(opts, 'install_type')
     mipData.install_type = opts.install_type;
 else
--- a/+mip/+resolve/detect_cpu_level.m
+++ b/+mip/+resolve/detect_cpu_level.m
@@ -1,0 +1,110 @@
+function level = detect_cpu_level()
+%DETECT_CPU_LEVEL   Detect the highest supported x86_64 SIMD level.
+%
+%   level = mip.resolve.detect_cpu_level()
+%
+%   Returns one of 'x86_64_v1', 'x86_64_v2', 'x86_64_v3', 'x86_64_v4',
+%   or '' on non-x86_64 platforms (macOS ARM, etc.).
+%
+%   Only relevant for linux_x86_64 and windows_x86_64 — the only platforms
+%   where SIMD variants are built.  Pure MATLAB, no MEX needed.
+
+arch = computer('arch');
+
+if strcmp(arch, 'glnxa64')
+    flags = get_linux_flags();
+elseif strcmp(arch, 'win64')
+    flags = get_windows_flags();
+else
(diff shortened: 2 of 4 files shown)


danfortunato:
This is the original PR from #109. It has not yet been reviewed by me.

magland:
Thanks for reopening this @danfortunato on the refactored mip code.

My main concerns are

(a) the windows cpu level detection, which doesn't look like something I would trust - I'm hoping there's a more direct way
(b) doesn't affect this PR, but on the mip-core/channel side, I don't like a boolean flag - I'd prefer being explicit about which levels to build.

@DiamonDinoia

DiamonDinoia:
(a) I don't have a windows machine to improve upon. 
(b) Why should it not be a boolean? What advantage does it give? I think we are the maintainers we should control the simd flags directly. The other maintainers opt in or out without having to worry about what these flags even mean. A lot of maintainers do -march=native and that's it.

magland:
> (a) I don't have a windows machine to improve upon.

Since we can't verify the Windows detection is reliable, I think we should leave Windows unsupported for now until we have a strategy. (I have a Windows 10 VM on my Linux box but it's painful to use — considering getting an actual Windows machine.)

> (b) Why should it not be a boolean? What advantage does it give?

Being explicit about which levels to build makes the mip.yaml self-documenting — a reader immediately understands what's being targeted rather than interpreting a mystery boolean. It also gives more flexibility if we ever need finer control. If we want to support a new level in the future we'd just update the mip.yaml files, which is some churn but corresponds to a real, intentional change.

One other concern (also about the channel side, not this PR directly): in my opinion, the environment variables from the runner shouldn't include cmake flags. It should be up to the package's compile.m to decide what to do with the target CPU information. For example, maybe a different build pipeline is being used (not cmake)

DiamonDinoia:
we can try CFLAGS CXXFLAGS but I never did it with cmake. I assume it respects them.

magland:
> we can try CFLAGS CXXFLAGS but I never did it with cmake. I assume it respects them.

Oh sorry, I was confused, I was thinking CFLAGS CXXFLAGS were cmake-specific.

In general I think the env var should specify the target version, not prescribe any flags.

What do you think @danfortunato ?

DiamonDinoia:
I am against this suggestion. I don't want the channel maintainer to have to convert simd level to compile flags.

I do not even trust that they know the correct flags.

magland:
> I am against this suggestion. I don't want the channel maintainer to have to convert simd level to compile flags.
> 
> I do not even trust that they know the correct flags.

The channel maintainer will first be us, and then it will be someone using our scripts as templates.

PR #136, opened 2026-04-10
Extract get_absolute_path to replace fragile what() calls

danfortunato:
## Summary
- Introduces `mip.paths.get_absolute_path`, a robust utility for resolving relative paths to absolute paths using `matlab.io.internal.filesystem.resolveRelativeLocation`
- Replaces all uses of MATLAB's `what()` for path resolution across `compute_addpaths`, `install_local`, and `bundle`, which was fragile and limited to directories on the MATLAB path
- Simplifies error handling by consolidating repeated `what()`/`isempty` checks into a single reusable function

### DIFF
--- a/+mip/+build/compute_addpaths.m
+++ b/+mip/+build/compute_addpaths.m
@@ -84,16 +84,8 @@ end
 
 function rel = getRelativePath(targetDir, baseDir)
 % Get relative path from baseDir to targetDir.
-    w = what(targetDir);
-    if isempty(w)
-        error('mip:build:notADirectory', '"%s" is not a directory.', targetDir);
-    end
-    targetDir = w.path;
-    w = what(baseDir);
-    if isempty(w)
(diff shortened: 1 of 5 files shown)


magland:
Numbl will need a shim for matlab.io.internal.filesystem.resolveRelativeLocation, so let's not merge this until I put that in place (numbl always pulls the latest mip as part of its operation at this point)

I assume there's not a standard builtin that will do this?

magland:
Actually, it looks like numbl's mip usage doesn't include any of these instances, so that's not a stopper. (in fact, numbl doesn't yet support "what" anyway)

danfortunato:
@magland I made another commit to use `fileattrib` instead of `matlab.io.filesystem...`: https://github.com/mip-org/mip/commit/9cdcb1699d50933b86c34d1e03b2186cd9f5e035 Is that better for numbl?

magland:
much better, thanks!

PR #137, opened 2026-04-10
Add --all and --deps flags to mip update

### LINKED ISSUE #133: Add `mip update --all`
Given that `mip update X` does not automatically update X's dependencies (see #99), I think we should support `mip update --all`

magland:
## Summary
- `mip update --all` updates every installed package
- `mip update --deps <pkg>` updates a package and its transitive dependencies
- Both flags can be combined with `--force`

Closes #133


+mip/update.m
```diff
@@ -423,3 +453,58 @@ function updateSelf(p, force)
     fprintf('\nmip has been updated to %s.\n', latestVersion);
 end
 
+function expanded = expandWithDeps(args)
```
danfortunato, line 461: Should this use `mip.resolve.get_all_dependencies`? Seems like possible duplicate code

PR #139, opened 2026-04-10
Use fileattrib for absolute path resolution

danfortunato:
## Summary

- Replace undocumented `matlab.io.internal.filesystem.resolveRelativeLocation` with the public `fileattrib` function for resolving absolute paths in `get_absolute_path`
- Simplifies the logic: `fileattrib` returns status and path info in one call, removing the need for a separate `exist` check


danfortunato:
@magland This replaces the `matlab.io.internal.filesystem.resolveRelativeLocation` shim with builtin `fileattrib` for numbl.

Issue #140, opened 2026-04-10
`mip update mip` is broken


danfortunato:
```
>> mip update mip
Checking for updates to mip (installed: main)...
Updating mip: main -> main
Downloading from https://github.com/mip-org/mip-core/releases/download/mip-main/mip-main-any.mhl...
Download complete.
Extracting package...
Warning: Removed
'/Users/dfortunato/Documents/MATLAB/mip/packages/mip-org/core/mip/mip'
from the MATLAB path for this MATLAB session.
	See 'doc path' for more information. 
> In [mip.update>updateSelf](matlab:matlab.lang.internal.introspective.errorDocCallback('mip.update>updateSelf', '/Users/dfortunato/Documents/MATLAB/mip/packages/mip-org/core/mip/mip/+mip/update.m', 404)) ([line 404](matlab: opentoline('/Users/dfortunato/Documents/MATLAB/mip/packages/mip-org/core/mip/mip/+mip/update.m',404,0)))
In [mip.update](matlab:matlab.lang.internal.introspective.errorDocCallback('mip.update', '/Users/dfortunato/Documents/MATLAB/mip/packages/mip-org/core/mip/mip/+mip/update.m', 67)) ([line 67](matlab: opentoline('/Users/dfortunato/Documents/MATLAB/mip/packages/mip-org/core/mip/mip/+mip/update.m',67,0)))
In [mip](matlab:matlab.lang.internal.introspective.errorDocCallback('mip', '/Users/dfortunato/Documents/MATLAB/mip/packages/mip-org/core/mip/mip/mip.m', 68)) ([line 68](matlab: opentoline('/Users/dfortunato/Documents/MATLAB/mip/packages/mip-org/core/mip/mip/mip.m',68,0))) 
Unrecognized function or variable 'latestVersion'.

Error in [mip/update>updateSelf](matlab:matlab.lang.internal.introspective.errorDocCallback('mip/update>updateSelf', '/Users/dfortunato/Documents/MATLAB/mip/packages/mip-org/core/mip/mip/+mip/update.m', 406)) ([line 406](matlab: opentoline('/Users/dfortunato/Documents/MATLAB/mip/packages/mip-org/core/mip/mip/+mip/update.m',406,0)))
        fprintf('Successfully updated mip to %s\n', latestVersion);
                                                    ^^^^^^^^^^^^^
[...]

PR #144, opened 2026-04-11
Add integration test for `mip update mip` self-update path

magland:
Adds `tests/TestUpdateSelf.m`, a remote-gated integration test covering the `updateSelf` path in `+mip/update.m`. It seeds a fake `mip-org/core/mip` under an isolated `MIP_ROOT`, calls `mip.update('--force', 'mip-org/core/mip')`, and verifies the swap completed against the real `mip-org/core` channel.

Wired into `run_tests.m` alongside the other `MIP_SKIP_REMOTE`-gated tests.

Would have caught #140 — verified by temporarily reintroducing the `latestVersion` reference, which reproduces the exact error from the issue.

_Composed by Claude, reviewed by @magland._


danfortunato:
Great!

PR #157, opened 2026-04-13
Convert source to char in download_mhl

danfortunato:
## Summary
- Add `source = char(source)` to handle MATLAB string inputs (not just char arrays)
- Clean up concatenation syntax and whitespace

## Test plan
- [ ] Verify `download_mhl` works when `source` is passed as a MATLAB string (`"https://..."`)
- [ ] Verify `download_mhl` still works with char array inputs (`'https://...'`)
- [ ] Verify local file copy path also handles string inputs

### DIFF
--- a/+mip/+channel/download_mhl.m
+++ b/+mip/+channel/download_mhl.m
@@ -17,6 +17,8 @@ if ~exist(destDir, 'dir')
     mkdir(destDir);
 end
 
+source = char(source);
+
 % Check if source is a URL or local file
 isURL = startsWith(source, {'http://', 'https://'});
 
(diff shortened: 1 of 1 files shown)


magland:
Which matlab operation, specifically, fails on string? I thought string/char were mostly interchangeable for operations.

danfortunato:
`['foo' 'bar']` concatenates chars into `'foobar'` but `["foo" "bar"]` does not (it is an array of strings). `"foo" + "bar"` concatenates into `"foobar"`.

danfortunato:
There might be other places we do this. I came across this awhile ago in this routine and had it changed locally but forgot to make an issue. Maybe we should go through all routines and make sure they consistently use chars/strings. We can support both at the top-level.

magland:
So you're suggesting we work with char throughout, except at top level?

danfortunato:
That would be my preference. I usually work with chars. The string syntax is newer for MATLAB

magland:
Okay sounds good. Should we merge this now, or wait for the larger PR? I'm okay either way.

danfortunato:
Let's merge this now and then make a new issue for the larger change

magland:
Sounds good.

Issue #91, opened 2026-04-07
Offer to replace existing mip install when already installed


magland:
@danfortunato when the install script detects that mip is already installed, it currently aborts. Instead, it should ask the user whether they want to replace the existing mip package directory. This is useful for recovering from a corrupted install.

magland:
Hmmm. I guess this gets complicated. Let me propose:

we provide a means of genuinely uninstalling mip... which would remove it from the path and clearing the entire mip directory (including packages). This is the responsible thing to do right? Maybe it would be uninstall.txt on the website? Then if the user had a broken state somehow, or wanted to start from scratch (or useful for us if we want to start from scratch), they could uninstall and then reinstall, both using the eval(webread(...)) method. Then this sidesteps the tricky parts of the present issue.

magland:
What I imagine:

If command mip exists, explain that mip is already installed. They can either do "mip update mip" or alternatively, following the instructions on mip.sh to completely uninstall. Then script ends - no prompting user.

If command mip does not exist, then we ask user for a root directory. If that directory already exists, then we exit explaining that the directory already exists.

Keep it simple like that.

magland:
> We could also probably figure out a way to have the uninstaller built into mip itself somehow.

I feel like since install is outside of mip package, uninstall should be outside the package as well.

danfortunato:
`mip uninstall mip` has been merged

PR #158, opened 2026-04-13
Add self-uninstall support for mip-org/core/mip

danfortunato:
## Summary

- `mip uninstall mip` (or `mip uninstall mip-org/core/mip`) now triggers a full self-uninstall flow: confirms with the user, resets state, removes mip from the saved MATLAB path, and deletes the root directory
- If the user declines the confirmation, other packages in the same command are still uninstalled (e.g., `mip uninstall mip-org/core/mip otherpkg` — declining mip removal still uninstalls `otherpkg`)
- Adds `MIP_CONFIRM` env var to bypass the interactive prompt in tests
- Moves self-uninstall tests from `TestUninstallPackage` into new `TestUninstallSelf` (8 tests)

## Test plan

- [x] All 377 tests pass (`run_tests`)


danfortunato:
This partially addresses #91. We still need to modify the install script as follows:

- If command mip exists, explain that mip is already installed. They can either do `mip update mip` or `mip uninstall mip`. Then script ends - no prompting user.
- If command mip does not exist, then we ask user for a root directory. If that directory already exists, then we exit explaining that the directory already exists.

+mip/uninstall.m
```diff
@@ -110,3 +108,74 @@ function uninstall(varargin)
...
+    end
+    didUninstall = true;
+
+    % Reset all loaded packages and key-value stores
+    mip.reset();
+
+    fprintf('Removing mip from saved MATLAB path...\n');
+
+    % Cache the user's current path
+    current_path = path;
```
magland, line 149: maybe there should be a try/catch between current_path = path and path(current_path) so that if something goes wrong, the user doesn't lose their current path.

magland (commented):
Looks good. Perhaps add try/catch as described... but I'm happy either way.

+mip/uninstall.m
danfortunato: Good idea. Done.

danfortunato: Should we have the same try/catch in install.txt when we are manipulating the path?

magland: Yes, I think that would make sense.

danfortunato: Implemented in https://github.com/mip-org/mip_website/pull/8

PR #159, opened 2026-04-13
Add code coverage support with Codecov integration

danfortunato:
## Summary
- Add optional `'Coverage', true` flag to `run_tests` that instruments all `+mip/` source files and generates Cobertura XML + HTML coverage reports
- Add a `coverage` CI job that uploads results to Codecov on every push/PR to main
- Add Codecov badge to README


magland:
LGTM

PR #160, opened 2026-04-13
Treat @ as literal in local paths, improve hint for @version

danfortunato:
## Summary
- Local paths (prefixed with `~`, `.`, `/`) now explicitly treat `@` as a literal path character, not a version separator. This supports MATLAB class directories (`@MyClass`) and directories with `@` in their names (`pkg@dev`).
- `buildLocalDirHint` now strips the `@version` suffix when checking for matching local directories. Previously, `mip install foo@1.0` with a local `foo/` directory showed no hint; now it suggests `mip install ./foo`. Exact-match directories (e.g., `foo@1.0/`) are still preferred.
- Added docs in §1.4 and §3.0 of behavior-reference.md describing `@` semantics in local paths.

Closes #107.

## Test plan
- [x] 5 new offline tests in `TestInstallLocal.m`: `./@foo`, `./pkg@1.0`, `/path/@foo` categorized as local; end-to-end copy and editable installs from `@`-containing directories
- [x] 2 new integration tests in `TestInstallRemote.m`: `foo@1.0` hint fires when `foo/` exists; exact-match `foo@1.0/` directory preferred over stripped `foo/`
- [x] All 378 tests pass

### DIFF
--- a/+mip/install.m
+++ b/+mip/install.m
@@ -25,6 +25,8 @@ function install(varargin)
 %   or a Windows drive letter (e.g. 'C:\path\mypkg', 'C:/path/mypkg').
 %   The directory must contain a mip.yaml file. In editable mode, changes
 %   to the source directory are reflected immediately without reinstalling.
+%   '@' in local paths is treated as a literal character, not a version
+%   separator (e.g. './@MyClass', './pkg@dev' are valid local paths).
 %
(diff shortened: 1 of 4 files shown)


magland:
It's hard for me to imagine a case where you'd have a mip package inside a directory with @, but this looks good to me.

danfortunato:
Yeah me too. But it can't hurt, right?

PR #173, opened 2026-04-15
Add `mip init` command for scaffolding a new package

magland:
## Summary
- New command `mip init <path> [--name <name>]` generates a starter `mip.yaml` in the given directory. Defaults the package name to the directory basename; `--name` overrides. Optional string fields (`description`, `version`, `license`, `homepage`, `repository`) are emitted blank so the user can see what to fill in. Build is `[any]`.
- Auto-discovers `addpaths` by walking the package directory and identifying folders with runtime MATLAB code (skips `tests/`, `docs/`, `build/`, platform/binary dirs, etc.).
- Creates a blank `test_<name>.m` (only if not already present — never overwrites) and references it from the generated `mip.yaml`'s `test_script` field.
- If `mip.yaml` already exists in the target dir, prints a message and exits without modifying anything.
- 18 new tests in `TestInit`. Full suite (382 tests) passes.

cc @danfortunato


magland:
`mip init` (no args) is now equivalent to `mip init .`.

PR #175, opened 2026-04-15
Support `mip install <zip-url> --name <name>`

magland:
Builds on #174 (which builds on #173). Once those merge to main, this base can be retargeted.

## Summary
- `mip install` now accepts an http(s) URL whose path ends in `.zip` (case-insensitive, ignoring query strings — so GitHub codeload URLs and presigned download URLs both work). Flow: download → extract → unwrap a single top-level subdir if present (handles `repo-main/` wrappers) → auto-init mip.yaml if missing → `install_local` (non-editable) → cleanup temp dir.
- `--name <name>` is required for zip URL installs and the URL is recorded as `repository:` in the auto-generated mip.yaml.
- `mip init` gains a `--repository <url>` option to fill in that field instead of leaving it blank.
- New error ids: `mip:install:nameRequiredForZipUrl`, `mip:install:nameRequiresZipUrl`, `mip:install:multipleZipUrls`, `mip:install:missingNameValue`, `mip:install:zipDownloadFailed`, `mip:install:zipExtractFailed`. `--editable` with a zip URL surfaces `mip:install:editableRequiresLocal`.
- 14 new tests in [TestInstallZipUrl.m](https://github.com/mip-org/mip/blob/feature/install-zip-url/tests/TestInstallZipUrl.m) cover flag validation, URL detection (query strings, GitHub-style, uppercase, `.zip` only in query string, `.mhl` not classified as zip), and download-error surfacing. 2 new tests in [TestInit.m](https://github.com/mip-org/mip/blob/feature/install-zip-url/tests/TestInit.m) cover the new `--repository` option. End-to-end download path verified via manual smoke test (downloaded `mip-org/mip-hello/archive/refs/heads/main.zip`, auto-init populated `repository`, package installed under `local/local/`).
- Behavior reference §3.4 added; error-id table updated.

## Known limitation
Because the source dir is deleted after install, `mip update` cannot reinstall from the original URL — the user must re-run `mip install <url> --name <name>`.

Full local suite (407 tests) passes.

cc @danfortunato


magland:
Oops — I want to change the CLI shape. Revising to `mip install <name> --url <zip-url>` (name as positional, URL as the flag) instead of `mip install <zip-url> --name <name>`. New commit incoming.

magland:
Closing in favor of fresh PR with revised CLI: `mip install <name> --url <zip-url>`.

PR #176, opened 2026-04-16
Support `mip install <name> --url <zip-url>`

magland:
Builds on #174 (which builds on #173). Once those merge to main, this base can be retargeted.

## Summary
- `mip install <name> --url <zip-url>` installs a package from a remote `.zip` archive, using the positional `<name>` as the package name. Flow: download → extract → unwrap a single top-level subdir if present (handles `repo-main/` GitHub-archive wrappers) → auto-init mip.yaml if missing (with `--repository <url>` so the source URL is recorded) → non-editable `install_local` → cleanup temp dir on success or failure.
- `--url` value must be an http(s) URL whose path (ignoring query/fragment) ends in `.zip`, case-insensitive.
- `mip init` gains a `--repository <url>` option to fill in that field instead of leaving it blank.
- Constraints: exactly one positional (bare name only — no FQN/URL/path), at most one `--url` per call, `--editable` rejected.
- New error ids: `mip:install:urlRequiresName`, `mip:install:urlTakesSingleName`, `mip:install:urlMustBeZip`, `mip:install:multipleUrls`, `mip:install:missingUrlValue`, `mip:install:zipDownloadFailed`, `mip:install:zipExtractFailed`.
- 17 new tests in [TestInstallZipUrl.m](https://github.com/mip-org/mip/blob/feature/install-url-flag/tests/TestInstallZipUrl.m) covering flag/positional validation, URL detection (query strings, GitHub-style, uppercase, `.zip` only in query string rejected, non-http rejected, `.mhl`/`.tar.gz` rejected), and download-error surfacing. 2 new tests in [TestInit.m](https://github.com/mip-org/mip/blob/feature/install-url-flag/tests/TestInit.m) cover `--repository`. End-to-end verified via manual smoke test: `mip install mip_hello_smoke3 --url https://github.com/mip-org/mip-hello/archive/refs/heads/main.zip` installed successfully with `repository:` populated to the URL.
- Behavior reference §3.4 added; error-id table updated.

## Known limitation
Because the source dir is deleted after install, `mip update` cannot reinstall from the original URL — the user must re-run `mip install <name> --url <zip-url>`.

Full local suite (412 tests) passes.

Supersedes #175 (which used `mip install <zip-url> --name <name>`).

cc @danfortunato


danfortunato:
What happens when a user types `mip update` and `mip update --force` in this case? Does it error or does nothing happen?

danfortunato:
We could store the md5 hash of the zip file when they install from a URL zip. `mip update` could then check the live URL zip hash against the stored hash to see if something changed. What do you think?

PR #177, opened 2026-04-16
Support installing from MathWorks File Exchange URLs

magland:
Builds on #176 (which builds on #174 → #173). Once those merge to main, this base can be retargeted.

## Summary
- `mip install <name> --url <url>` now accepts a MathWorks File Exchange landing URL (anything starting with `https://www.mathworks.com/matlabcentral/fileexchange/`) in addition to a direct `.zip` URL.
- The landing URL is auto-resolved to the underlying download:
  1. Append `?download=true` (or `&download=true` if a query string is already present).
  2. HEAD the URL via `matlab.net.http.RequestMessage` with `User-Agent: curl/8.0` — the MathWorks Akamai layer returns 403 to MATLAB's default UA but accepts curl-style UAs.
  3. Follow the 302 redirect to the UUID-based `mlc-downloads/.../<file>.zip?src=&license=` URL.
  4. Strip the query string. The result is the canonical, stable zip URL.
- The resolved zip URL is both downloaded **and** recorded in the auto-generated mip.yaml's `repository` field, so the package retains a stable identifier.
- New error: `mip:install:fxResolveFailed` for network errors, non-2xx responses, or final URLs that don't end in `.zip`.

## Tests
- 1 validation test asserts an FX-shaped URL is **not** rejected by `mip:install:urlMustBeZip`.
- 1 end-to-end test (gated by `MIP_SKIP_REMOTE`) installs `shadedErrorBar` (FEX 26311 — small, widely-used, benign plotting utility) and verifies the recorded `repository:` ends in `.zip`, contains `mlc-downloads`, and has no query string.

Manual smoke tested with the same FX URL: install succeeded and `repository:` was populated to `https://www.mathworks.com/matlabcentral/mlc-downloads/downloads/.../packages/zip/raacampbell-shadedErrorBar-aa6d919.zip`.

Full local suite (414 tests) passes; FX e2e test passes when remote is enabled (19/19 in TestInstallZipUrl).

cc @danfortunato


danfortunato:
We could probably support versioning and `mip update` for FEX packages when the URL is not a direct .zip:

- File Exchange keeps a version history and we could see if the installed version is older than the latest version from File Exchange when `mip update` is fun. When we do `mip.init(sourceDir, '--name', pkgName, '--repository', zipUrl);` we could also keep track of the FEX version number for tracking purposes.
- `mip install <name>@<version> --url <fex_url>` could use that version from FEX if it exists and error if not.

PR #182, opened 2026-04-17
Reject .mhl archives with path-traversal entries

### LINKED ISSUE #163: Path traversal in .mhl extraction
`+mip/+channel/extract_mhl.m` calls `unzip()` directly with no validation that extracted paths stay under `destDir`. A malicious `.mhl` with entries like `../../foo` can overwrite files outside the package directory.

Since `mip install <url>` and `mip install file.mhl` accept arbitrary archives, this is exploitable.

Fix: after `unzip`, verify each returned path (and any symlinks) resolve under `destDir`. Reject the archive on violation.

magland:
## Summary

Fixes #163. MATLAB's `unzip()` extracts ZIP entries as-is — a malicious `.mhl` with `../` entries, or one mis-marked as a symlink, can write files anywhere the MATLAB process has permission. This PR adds a pre-extraction validator that rejects unsafe archives before any bytes touch disk.

## What changed

**`+mip/+channel/assert_mhl_safe.m`** (new) — Parses the archive's End of Central Directory, each Central Directory entry, and each Local File Header in pure MATLAB. Errors with `mip:pathTraversal` on any unsafe entry.

**`+mip/+channel/extract_mhl.m`** — Calls `assert_mhl_safe` before `unzip()`.

**`+mip/install.m` (`downloadAndInstall`)** — Extracts into a `staging/` subdir of the temp dir and then `movefile`s to `pkgDir`, matching the pattern used by the other callers. Keeps rollback clean on `movefile` failure.

## What the validator rejects

Entry names — checked in both the central directory and each local file header (see "Design decisions" below):

- Absolute paths (leading `/`) and Windows drive letters (`C:...`)
- Null bytes in names
- `..` components that resolve *outside* the extraction root. Benign in-tree `..` like `foo/../bar` is permitted.
- CD/local-header name disagreements (a ZIP spec violation)

Symlink entries (Unix `S_IFLNK` in the high 16 bits of the CD external file attributes):

- Target resolved relative to `dirname(link)` must stay inside `destDir`; benign in-tree symlinks like `sub/link → ../other` are permitted.
- Compressed symlink payloads are rejected as a safe fallback (cannot read target without a deflate implementation; compressed symlinks are vanishingly rare in practice).

## Design decisions

- **Pre-extraction rather than post-extraction.** An earlier draft of this PR did a post-`unzip()` prefix check. That detected violations only *after* writes occurred — `unzip()` follows `..` freely, so a malicious archive could clobber `~/.ssh/authorized_keys`, `startup.m`, etc. before the check fired. Staging under `/tmp` didn't contain the damage because the `..` escapes `/tmp` entirely. Pre-extraction validation means no bytes hit disk if the archive is unsafe.

- **Validate local-header names, not just central-directory names.** Empirically verified (with a hand-crafted mismatched archive) that MATLAB's `unzip()` honors the local file header name. A check that only reads the central directory is bypassable by an attacker who puts safe names in the CD and malicious names in local headers. The validator reads both and rejects any disagreement.

- **Pure MATLAB, no Java/Python.** A small hand-written ZIP parser (~50 lines of central-directory + local-header walking) avoids a Java dependency and keeps test fixtures trivially cross-platform. Earlier draft test fixtures used Python and broke on Windows CI.

- **Targeted symlink check rather than blanket rejection.** Only symlinks whose resolved target escapes `destDir` are rejected, so legitimate in-tree symlinks (rare but possible in packages bundling `.framework` directories etc.) still extract normally.

## Tests

`tests/TestExtractMhl.m` — 13 cases:

- Valid archive extracts
- Path traversal: shallow `../x`, deep `a/b/../../../x`, absolute `/etc/...`, Windows drive `C:\...`, backslash `..\x`
- Benign in-tree `..` accepted
- Missing `mip.json` rejected
- CD vs local-header name mismatch rejected
- Symlinks: absolute target rejected, escaping relative target rejected, benign relative target accepted
- Non-ZIP input rejected

ZIP fixtures are hand-assembled in pure MATLAB with a small CRC-32 helper — no Python, no Java, cross-platform.


magland:
CI was failing on Windows because the `TestExtractMhl` helpers built the test `.mhl` fixtures by shelling out to `python3 -c "..."`. The shell quoting and the `python3`-vs-`python` name difference don't work on Windows runners. Removing the test for now — rebuilding those fixtures in pure MATLAB isn't clean (MATLAB's `zip()` can't write a `../escaped.txt` entry without going through Java or shipping a binary fixture), and all three approaches considered were worse than just dropping the test. The guard itself is still exercised by the existing remote install tests.

danfortunato:
I reworked the approach from post-extraction checking to pre-extraction validation in 3d4b213. The key empirical finding: MATLAB's `unzip()` honors the local file header name, not the central directory name (verified with a hand-crafted mismatched archive). That made the original post-check problematic in two ways — it couldn't prevent writes, only detect them, and a CD-only check would have been bypassable.

Also updated the PR description to describe the current design rather than the original post-extraction approach. Commit log (`4ccba5b`, `0399dfa`, `3d4b213`) preserves the full evolution. Happy to iterate if you'd prefer a different direction.

PR #183, opened 2026-04-17
Atomic write for directly_installed.txt

### LINKED ISSUE #164: Non-atomic write of directly_installed.txt can lose state
`+mip/+state/set_directly_installed.m` opens the file with `'w'` and writes entries line-by-line. A crash, Ctrl-C, or disk-full between `fopen` and the final `fprintf` leaves the file truncated or empty.

Consequence: the next `mip uninstall` or post-install prune sees an incomplete root set and deletes packages that were actually directly installed, treating them as orphans.

Fix: write to `directly_installed.txt.tmp` and rename into place.

magland:
Closes #164.

## Summary
- Write package list to `directly_installed.txt.tmp` first, then `movefile` into place, so a crash or Ctrl-C mid-write can no longer leave the file truncated.
- On failure, the tmp file is cleaned up and the original `directly_installed.txt` is left untouched.

### DIFF
--- a/+mip/+state/set_directly_installed.m
+++ b/+mip/+state/set_directly_installed.m
@@ -5,22 +5,22 @@ function set_directly_installed(packages)
 %   packages - Cell array of package names that are directly installed
 
     packagesDir = mip.paths.get_packages_dir();
-    
-    % Create packages directory if it doesn't exist
+
     if ~exist(packagesDir, 'dir')
         mkdir(packagesDir);
     end
-    
+
     directFile = fullfile(packagesDir, 'directly_installed.txt');
-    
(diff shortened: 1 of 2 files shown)


magland:
Added `testDirectlyInstalled_StaleTmpReplaced` in `TestPackageDiscovery` covering the happy path: a leftover `directly_installed.txt.tmp` from a prior crash must be replaced (not appended to), and must not survive a successful write.

Not adding tests for the two error branches (`fopen` on the tmp file fails, `movefile` fails), since simulating disk-full / permission errors portably across Linux / macOS / Windows would be more infrastructure than it's worth for two error paths that just rethrow with a clearer message.

PR #184, opened 2026-04-17
Don't mark .mhl dependencies as directly installed

### LINKED ISSUE #165: .mhl install marks dependencies as directly installed
In `+mip/install.m`, `installFromMhl` calls `installFromRepository(pkgInfo.dependencies, channel)` to pull in the package's dependencies. But `installFromRepository` treats its argument list as user-requested packages and appends every entry to `directly_installed.txt` (see the loop at the end of that function).

Effect: dependencies of an `.mhl` install are permanently marked as directly installed and will never be pruned, even after the parent package is uninstalled.

Fix: install `.mhl` dependencies through a path that does not add them to `directly_installed.txt` (e.g. a private helper that skips the "directly installed" marking).

magland:
Closes #165.

## Summary
- `installFromRepository` now takes a `markDirectlyInstalled` flag (default `true`).
- `installFromMhl` passes `false` when pulling in its package's dependencies, so those deps stay prunable when the parent `.mhl` package is later uninstalled.


magland:
codecov/patch is red on the new `if markDirectlyInstalled` branch. Leaving it uncovered intentionally:

- Exercising it requires building a `.mhl` archive with declared dependencies and then having `installFromMhl` resolve those deps against a real remote channel. That's significant test infrastructure for a single one-line guard, and building `.mhl` fixtures portably across Linux / macOS / Windows is the exact problem we just hit in #182.
- `installFromRepository` is a local function inside `+mip/install.m`, so it can't be unit-tested in isolation without refactoring it into `+mip/+internal/` — out of scope for this fix.

The diff is small and obviously correct on inspection; flagging rather than padding the PR.

PR #203, opened 2026-04-20
Move package path list from scripts into mip.json

magland:
## Summary

- Bundled / installed packages now declare their addpath list in `mip.json` under a new `paths` field instead of generated `load_package.m` / `unload_package.m` files.
- `mip.load` reads `paths`, `addpath`s each entry against the source dir (via `mip.paths.get_source_dir`), then runs `load_package.m` if present. `mip.unload` mirrors this: legacy `unload_package.m` runs first, then `rmpath` the `paths` entries, then the existing defensive sweep.
- Bundle and local install (editable and copy) no longer emit `load_package.m` / `unload_package.m`.

## Compatibility

Legacy installs (no `paths` field, scripts present) keep working unchanged — the scripts still fire. `mip:loadNotFound` / `mip:unloadNotFound` only surface when **neither** a `paths` field nor the corresponding script exists. `mip.update` self-update resolves old + new path lists from the respective `mip.json`s before swapping the installed copy, so it works whether either side is legacy or new-style.

## Rollout note

Bundles produced by this version require a mip client that understands the `paths` field. An older mip that extracts a new-style bundle will find no `load_package.m` and raise `mip:loadNotFound`. Users should update mip before installing packages rebuilt under this scheme.


magland:
@danfortunato we won't really see if this is working until after this is merged and mip package gets rebuild and then we rebuild some packages. If it looks reasonable I'd say we should go ahead and merge and then run the mip-core, and after that goes through, we increment the release numbers of a couple packages. Then we mip install them and verify the load_package.m is gone.

PR #213, opened 2026-04-22
Use reserved '_' org for non-channel packages

### LINKED ISSUE #197: figure out how to name local/local, local/fex, etc.

magland:
## Summary

Resolves #197. Non-channel packages (local directory installs, File Exchange/URL installs) now live under a reserved `_` org with a subchannel identifying the install source:

- `_/local/<pkg>` — directory and editable installs (previously `local/local/<pkg>`)
- `_/fex/<pkg>` — File Exchange and `--url` installs (previously lumped in with `local/local/<pkg>`)

**User-facing display strips the `_/` prefix**, so `mip list`, `mip info`, install/load/update messages, and the "to use this package, run: mip load X" hint all show `local/<pkg>` and `fex/<pkg>`.

**Users can type either form.** The parser accepts a 2-part input `category/name` as shorthand for the internal FQN `_/category/name`, so `mip load local/mypkg` and `mip load _/local/mypkg` are equivalent. Bare names (`mip load mypkg`) still work when unambiguous.

Bare-name resolution priority is unchanged: `mip-org/core` first, then alphabetical by FQN (with `_/` entries participating).

## Notable changes

- New helper `mip.parse.display_fqn` strips the `_/` prefix for display.
- `parse_package_arg` now accepts 2-part input (`X/Y` → `_/X/Y`).
- `install_local` takes an optional `subchannel` arg; the URL/FEX flow in `install.m` passes `'fex'`.
- `update.m` treats any `_/*` package as local-style (no channel lookup). Cleanup paths are derived from the package dir rather than hard-coded.
- No migration: existing on-disk state from the old `local/local/` layout is not converted. Start from a clean root.


magland:
@danfortunato this one is kind of big. We can go through it together if you want.

magland:
We decided we're going to go with a different system. We will restructure the fqn's and how installed packages are stored on disk

mip_root/packages/gh/mip-org/core/pkg1
mip_root/packages/gh/mip-org/staging/pkg2
mip_root/packages/gh/user/chan/pkg3
mip_root/packages/local/pkg4
mip_root/packages/fex/pkg5

the fqn's would be

gh/mip-org/core/pkg1
local/pkg4
fex/pkg5

so we'll have variable-length fqn's

For the github cases, we want the user to be able to omit the gh/ prefix whenever inputting a package name

And when displaying, always omit the gh/ prefix

but internally, it should be included in the fqn

As before, when unambiguous the last component of a fqn can be used as input from the user

So a new PR should be made along these lines to replace this one.

magland:
Replaced by #215, which takes a different approach: rather than a reserved `_` org, FQNs now carry a source-type prefix (`gh/`, `local/`, `fex/`) and become variable-length.

Issue #211, opened 2026-04-21
Move `version` from mip.yaml to recipe.yaml (or override in recipe.yaml)


danfortunato:

magland:
I already did this, because it's a change in the channels, not in this repo. See https://github.com/mip-org/mip-core/tree/main/packages/mip/numbl

danfortunato:
So `version` appears in both mip.yaml and recipe.yaml? Is that how it should be?

Should we migrate all packages to have version stored in recipe.yaml?

magland:
What we decided:

`version` is not an allowed field of recipe.yaml and it is an optional field of mip.yaml. If missing, that's equivalent to an empty version.

In the channel, the version is determined by the directory... so `packages/pkg1/main/recipe.yaml` would be version main, and `packages/pkg2/0.5.3/recipe.yaml` would be version 0.5.3.

The mip.yaml would never have version="main". Instead the mip.yaml version would either be blank or a numeric version.

Here are the rules for an allowed version (version directory name) in a channel. The version name (version directory name) must be one of the following
* The numeric version in the mip.yaml
* The name of the `branch` field in the recipe.yaml
* Any string if the mip.yaml has an empty or missing `version` field.

Does this sound right, @danfortunato ?

Issue #191, opened 2026-04-17
Figure out what to do about empty/`"unknown"` versions


danfortunato:

magland:
Dan and I decided: there won't ever be "unknown" of "unspecified". It will just be empty, and for display purposes (like in mip list) it will just show an empty string.

PR #218, opened 2026-04-23
Make mip.yaml version optional; enforce blank-or-numeric

magland:
Resolves the decision in #211. Related: #191.

## Summary

- `version` is optional in `mip.yaml`. If present, it must be blank or numeric.
- Non-numeric values (branch names like `main`/`numbl`) belong in the channel release-directory name, not in `mip.yaml`.
- `mip/mip.yaml` now has `version: ""` (was `"main"`, which is no longer allowed).
- `mip init` emits `version: ""`; `read_mip_yaml` normalizes empty/missing to `"unknown"` for internal display (addressing #191).
- Spec §1.3, §9.4, §11.2, and new §11.2.1 document the channel version rules.

## Coordinated merge

This change must land together with:
- mip-org/mip-channel-template#7
- mip-org/mip-core#5

Needs approval from @danfortunato.

## Follow-up

After merging, the `numbl` tag should be retagged to include this commit (the current `numbl` tag is behind main and has the old `version: "main"` in `mip.yaml`, which the new `prepare_packages.py` will reject).

### DIFF
--- a/+mip/+build/create_mip_json.m
+++ b/+mip/+build/create_mip_json.m
@@ -6,6 +6,7 @@ function create_mip_json(outputDir, mipConfig, architecture, opts)
 %   mipConfig      - Struct from read_mip_yaml
 %   architecture   - Effective architecture string
 %   opts           - (Optional) Struct with fields:
+%     .version     - version string to record (overrides mipConfig.version)
 %     .editable    - true for editable installs
 %     .source_path - original source path (for editable installs)
 %     .install_type - install type string (default: 'local')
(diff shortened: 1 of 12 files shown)


danfortunato:
Update following review discussion — pending in the local branch, will push shortly:

1. **No more `"unknown"` sentinel.** `read_mip_yaml`, `read_package_json`, `list`, and `update` all default a missing/blank version to `''` instead of `"unknown"`. Tests updated.

2. **`mip.yaml` in the bundle is no longer mutated by the channel build.** Instead, `prepare_packages.py` drops a `.release_version` side file alongside `.source_hash` / `.commit_hash`. On the MATLAB side:
   - `prepare_package.m` reads `.release_version` and passes it to `create_mip_json.m` via `jsonOpts.version`.
   - `create_mip_json.m` records it as `mipData.version`, overriding `mipConfig.version`.
   - `bundle.m` reads the bundled `mip.json`'s version for the `.mhl` filename.
   - The "Preparing package ... (version X)" status line uses the effective version too.

Net effect: the `mip.yaml` inside the `.mhl` is exactly as authored in the source repo, while `mip.json` carries the authoritative release-dir version. Spec §11.2.1 updated accordingly.

PR #219, opened 2026-04-23
Add extra_paths and mip load --with <group>

### LINKED ISSUE #199: Add --with-examples / --with-tests flags to mip load
## Summary

Channels currently exclude `examples/`, `demos/`, `tutorials/`, and `tests/` directories from package addpaths by convention (see [`+mip/+init/auto_add_paths.m`](https://github.com/mip-org/mip/blob/main/%2Bmip/%2Binit/auto_add_paths.m) for the canonical skip list, and consistent practice across chunkie, chebfun, gptoolbox, flam, kdtree, distmesh, etc.). This keeps the user's path clean and avoids shadowing.

The downside is friction for users who *do* want to run an upstream example: they have to `run(fullfile(mip.info('distmesh').loadPath, 'examples', 'ex01_unit_circle'))` or `cd` to the directory, instead of just `ex01_unit_circle`.

Proposed: opt-in flags on `mip load` that keep the default clean but make examples/tests easy to reach when asked for.

## Proposed UX

```matlab
mip load distmesh                    % current behavior — only src/ on path
mip load --with-examples distmesh    % also addpath the directories declared in example_paths:
mip load --with-tests distmesh       % also addpath the directories declared in test_paths:
mip load --with-examples --with-tests distmesh
```

## Implementation sketch

- **Flag parsing:** `mip load` accepts `--with-examples` and `--with-tests`.
- **Declared, not heuristic.** `mip.yaml` gains two optional fields that mirror `addpaths:`:

    ```yaml
    addpaths:
      - path: "src"

    example_paths:
      - path: "examples"

    test_paths:
      - path: "tests"
    ```

  When the user passes `--with-examples`, mip adds the entries in `example_paths:` to the MATLAB path alongside the normal `addpaths:`. Same for `--with-tests` / `test_paths:`. This matches how `addpaths:` already works (declared in the manifest, not detected at load time), so it's one consistent mental model rather than two.
- **No heuristic directory detection.** At load time we already have a hand-authored manifest — picking up on dir names by convention is extra magic with no corresponding benefit, and it can get degenerate cases wrong (e.g. a project where `examples/` actually contains test fixtures, or a package like `find-poly` whose subdirectory names happen to collide with skip-list entries for unrelated reasons).
- **Unload bookkeeping:** mip already tracks which paths it added at load time. The extras go in the same tracking list, so `mip unload` removes them alongside the rest. No new mechanism needed.
- **Flag with no declaration:** `mip load --with-examples <pkg>` on a package that didn't declare `example_paths:` produces a no-op with a warning like `no example_paths: configured in mip.yaml — ignoring --with-examples`. More predictable than silently scanning and possibly finding nothing.

## Open questions

- **Naming:** `--with-examples` reads naturally but is long. Short aliases (`--examples`, `-E`/`-T`)? I lean toward the long form for clarity.
- **Stickiness:** should there be a mechanism to make `--with-examples` persistent across sessions for a given package (e.g. a setting stored with the install record)? I'd say no for v1 — keep it explicit per-load. Easy to add later if users ask.
- **Combined flag:** `--with-extras` that turns on both? Probably unnecessary — typing two flags is fine.

## Motivation / prior discussion

Came up while adding the `distmesh` package to `mip-staging`: upstream's `startup.m` adds both `src/` and `examples/`, and the README's Quick Start is literally "copy-paste an example" — which silently fails under the exclude-examples-by-default convention. That prompted a broader look at the convention. See [`mip-staging:packages/distmesh/releases/main/`](https://github.com/mip-org/mip-staging/tree/main/packages/distmesh/releases/main) and [`auto_add_paths.m`](https://github.com/mip-org/mip/blob/main/%2Bmip/%2Binit/auto_add_paths.m) for the current state.

danfortunato:
## Summary
- Packages can declare opt-in path groups under `extra_paths:` in mip.yaml (examples, tests, benchmarks, or any custom name); `mip load --with <group>` adds those groups' directories to the MATLAB path at load time.
- Renames the top-level mip.yaml field from `addpaths:` to `paths:` to align with the existing `paths` field in mip.json.
- `mip init`'s auto_add_paths now routes top-level tests/, examples/, and benchmarks/ dirs (and their variants) into the generated `extra_paths` section instead of skipping them; `mip info` displays the declared groups. Warns once per load if no loaded package declared a requested group.

Resolves #199.


danfortunato:
@magland I wasn't sure if we decided on `--with-examples` or `--with examples`

magland:
I prefer --with examples

This looks good to me except for one thing:

Right now this is a breaking change for all existing packages on channels as well as currently installed packages. I suggest we make addpaths a synonym for paths in the mip.yaml, and produce a deprecation warning. Once we are sure the old packages have cleared out, then we switch to an error on the addpaths field.

PR #225, opened 2026-04-29
Fix pruning of explicitly re-installed transitive dependencies

### LINKED ISSUE #224: Pruning uninstalls a directly installed package when it was already installed as a dependency
I think the pruning behavior is not correct. We aren't marking things as "directly installed" when they should be:
```
mip install surfacefun   % also installs chebfun
mip install chebfun      % chebfun is already installed. should be marked as "directly installed" now
mip uninstall surfacefun % prunes chebfun!
```

Similarly:
```
mip install surfacefun   % also installs chebfun
mip load chebfun         % chebfun is already installed but now the user is directly interacting with it
                         % should it be considered "directly installed" now so it is not pruned?
mip uninstall surfacefun % prunes chebfun!
```

magland:
Fixes #224.

When `mip install <pkg>` is called for a package that's already on disk as a transitive dependency, the package wasn't being marked as `directly_installed`. The next `mip uninstall` of its parent then pruned it.

Two call sites had this shape:

- `installFromRepository` ([+mip/install.m](+mip/install.m)): the `add_directly_installed` loop was inside `if ~isempty(toInstallFqns)`, so it never ran when the requested package was already on disk. Hoisted it out, kept it after the install try/catch so a mid-install failure still rolls back orphans correctly.
- `installFromMhl` ([+mip/install.m](+mip/install.m)): the early-return "already installed" path skipped marking. Added `add_directly_installed` before the return.

The `mip load chebfun` variant in the issue is left unchanged — `load` doesn't currently touch `directly_installed`, and the issue marks that case as an open question.


magland (commented):
**Review by Claude (Anthropic)** — automated review on behalf of @magland.

Reviewed the diff against the failure modes of `directly_installed` tracking. No correctness issues found:

- Hoisted block in `installFromRepository` runs after the install try/catch (so a mid-install rethrow skips marking), and runs unconditionally when nothing needed downloading (which is the bug fix).
- `installFromMhl` early-return is the symmetric fix for the mhl path.
- `add_directly_installed` is idempotent, so re-marking is safe.
- `replaceExistingVersions` removes the mark before rebuilding; the new hoisted block re-adds it. No regression.
- Tests exercise the end-to-end scenario from #224.

LGTM.

PR #236, opened 2026-05-01
Print help when installing FEX URL without name

### LINKED ISSUE #235: Installing from FEX with no name should print a help message

magland:
Closes #235.

`mip install <FEX URL>` (without `--url`) used to fall into the .mhl
download path and fail with a generic download error. It now produces
the help message requested in the issue:

```
To install a package from the file exchange, you must specify a package name using the syntax
   mip install <name> --url <url>
```

Detection reuses the existing `isFileExchangeUrl` helper, so it covers
the same `https://www.mathworks.com/matlabcentral/fileexchange/...` (and
`http://`) prefixes that `--url` already auto-resolves.


magland:

Small, focused fix. The `isFileExchangeUrl` check is placed at the right point — inside the existing `endsWith('.mhl') || startsWith('http://') || startsWith('https://')` arm in argument classification, before the URL is enqueued for the `.mhl` download path, so the error fires before any network activity.

The helper is the same one used at line 678 to recognize FEX URLs in `--url` resolution, so the detection is consistent across both code paths.

Test covers the meaningful case — a real-shape FEX URL passed as a positional arg raises `mip:install:fexRequiresName`.

No serious problems found.

PR #237, opened 2026-05-01
Use CLI-style usage examples in subcommand docstrings

### LINKED ISSUE #234: Change subcommand docstrings to use spaces, etc

magland:
## Summary
- Replace MATLAB function-call syntax (`mip.install('pkg')`) with the shell/CLI form (`mip install <package>`) across the `+mip/*.m` subcommand docstrings.
- Matches the style already used in `mip.m` and how users actually invoke the commands.

Closes #234.


magland:

Pure docstring change — only the `% Usage:` blocks are modified, no function bodies touched. The CLI form is consistent with how the commands are actually invoked through `mip.m`'s dispatcher.

One minor observation, not a problem: `mip.index` is technically a return-valued function (`url = mip.index(...)`) and the new docstring documents only the CLI form (`mip index --channel ...`), since the dispatcher in `mip.m:159-161` parses `--channel` and prints the result. Users calling `mip.index` directly from MATLAB would need to use the positional form `mip.index('owner/channel')`, which is no longer documented. Acceptable trade-off if the goal is CLI-only docstrings.

Will conflict trivially with #242 in `pin.m`/`unpin.m` (#242 rewrites the prose paragraph below Usage; #237 changes only the Usage block) — easy to resolve.

No serious problems found.

PR #239, opened 2026-05-01
Stage @version replace to backup; restore on failure (#232)

### LINKED ISSUE #232: mip install foo@v2 destroys the old version before the new download starts
`+mip/install.m:477-478` (`replaceExistingVersions`):

```matlab
rmdir(pkgDir, 's');
mip.state.remove_directly_installed(s.fqn);
```

Runs before `downloadAndInstall`. If the download fails, foo is gone and no longer tracked, so the catch's `prune_unused_packages` can't restore it. User has to reinstall from scratch.

Fix: stage to a backup dir, restore on failure (same pattern as `downloadAndReplace`).

magland:
Fixes #232.

`mip install foo@v2` used to `rmdir` the existing `foo` and drop it from `directly_installed.txt` before the new download started. A download failure left the user with no copy of `foo` and no way for the catch's `prune_unused_packages` to restore it.

## Changes

- `replaceExistingVersions` moves the old package to a temp backup dir (instead of deleting it) and returns the list of backups, including whether the package was directly installed.
- New `restoreReplacementBackups` / `cleanupReplacementBackups` helpers restore on failure and discard on success.
- The install try/catch was widened to cover the equivalent-name check so it can't strand a backup either. `prune_unused_packages` is now gated on an `installAttempted` flag so the equivalent-name path doesn't accidentally delete unrelated installed packages. Restore runs before prune so transitive deps of the restored package survive.
- New `tests/TestInstallVersionRollback.m` writes a synthetic channel index with an unreachable `mhl_url` and verifies the old version is restored on disk and still tracked as directly installed.


magland:

The move-to-backup approach correctly fixes the data loss: `replaceExistingVersions` now stages old packages to temp dirs before the download begins, and `restoreReplacementBackups` moves them back on failure.

A few things worth checking:

**Order of operations in the catch block is correct.** `restoreReplacementBackups` runs before `prune_unused_packages`, so the restored package is back in `directly_installed.txt` before prune decides what to keep. Dependencies of the restored package won't be pruned. ✓

**`installAttempted` flag is correct.** It's set inside `if ~isempty(toInstallFqns)` after printing the plan but before the download loop, so the equivalent-name check (which fires before any download) doesn't trigger a prune, while a mid-download failure does. ✓

**Partial restore failure is handled.** `restoreReplacementBackups` catches restore errors per-backup and warns rather than throwing, so one failed restore doesn't skip the others.

**One small gap** (pre-existing, not introduced by this PR): in `replaceExistingVersions`, `remove_directly_installed` is called before `movefile`. If `movefile` throws, the backup struct isn't appended and the caller's catch has no backup to restore — the package dir is still on disk but no longer tracked as directly installed. This was equally broken in the old `rmdir` path. Not a regression.

Test covers the core scenario end-to-end: bad URL → download fails → old version present → directly_installed intact.

No serious problems found.

PR #240, opened 2026-05-01
Reject path-traversal entries in mip.json paths and --addpath/--rmpath

### LINKED ISSUE #230: paths / extra_paths from mip.json are not bounded to the package source dir
`+mip/load.m:269-286` does:

```matlab
target = fullfile(srcDir, rel);
addpath(target);
```

with no check that `rel` stays inside `srcDir`. A package whose `mip.json` ships `"paths": ["../../../.."]` will addpath an arbitrary ancestor on `mip load`, exposing sibling packages and shadowing user functions. Same for `extra_paths` and the `--addpath`/`--rmpath` handlers in the same file.

In isolation this needs an already-installed malicious package, but it is the amplifier on any archive-substitution attack — an attacker who can rewrite a single `.mhl` doesn't need MEX execution; they just point `paths` at `~`.

Proposed fix: validate each entry the same way `assert_mhl_safe` validates ZIP entries — reject `..` components, leading `/`, and Windows drive letters. Apply at install time (in `read_package_json` or `prepare_package`) and at load time as defense in depth.

magland:
## Summary
- A package whose `mip.json` declared `"paths": ["../../../.."]` would `addpath` an arbitrary ancestor at load time, exposing sibling packages and shadowing user functions. Same gap existed for `extra_paths` and the `--addpath` / `--rmpath` flags.
- Adds `mip.paths.assert_safe_relative` (mirrors the entry-name check in `assert_mhl_safe`): rejects null bytes, absolute paths, Windows drive letters, and `..` components that escape the source dir.
- Wired into `read_package_json` (validates `paths` and `extra_paths` on read) and into `load.m` (`applyMipJsonPaths`, `applyExtraPaths`, `applyPathAdjustments`) as defense in depth. Lets `mip:unsafePath` propagate through the dependency-load `try/catch` so it cannot fall through to `addpath`.

Closes #230


magland:

`assert_safe_relative`'s depth-tracking logic is correct: starts at 0, increments on real components, decrements on `..`, and errors the moment depth would go negative — so `foo/../..` and `../../etc` are both caught, but `foo/../bar` (depth stays ≥ 0 throughout) is allowed.

The two-layer defense is intentional and appropriate:

- `read_package_json` validates `paths` and `extra_paths` on parse. The `mip:unsafePath` is explicitly rethrown in `loadSingle`'s `try/catch` (which would otherwise downgrade it to a warning), so a malicious package cannot silently fall through to `addpath`.
- The duplicate checks in `applyMipJsonPaths`, `applyExtraPaths`, and `applyPathAdjustments` serve as defense-in-depth and as the *primary* guard for `--addpath`/`--rmpath` (those bypass `read_package_json` entirely).

Test coverage: the integration test constructs a real mip.json with `"paths":["../../../../etc"]` and verifies `mip:unsafePath` propagates from `mip.load`. The `--addpath` escape test covers the CLI flag path. Tests for `assert_safe_relative` itself would be a nice addition but aren't strictly required given the integration coverage.

No serious problems found.

Issue #201, opened 2026-04-20
SHA-256 mismatch when channel index is stale relative to published .mhl


magland:
Channel publishing is not atomic: the `.mhl` asset on a GitHub Release is updated (via `gh release upload --clobber`) before the `index.json` on GitHub Pages is refreshed by the `assemble-index` + `deploy` jobs. Any client that fetches the index during that window gets the old SHA-256 but the new `.mhl`, and `download_mhl` raises `mip:digestMismatch`.

Reproduced in CI: https://github.com/mip-org/mip/actions/runs/24680148850/job/72174904748 — `TestUpdateSelf/testUpdateSelf_ForceReplacesFakeWithReal` failed with expected `42fab336…` but actual `56a360e5…` for `mip-main-any.mhl`. The test itself is fine; it just happened to hit the inconsistency window of a concurrent `mip-core` rebuild.

The window exists on every `main` push to a channel repo: platform jobs upload clobbered assets in parallel, and `assemble-index`/`deploy` only run after all of them finish. Floating tags like `mip-main` are hit hardest because they're republished on every push.

Possible fixes:
- Content-addressed asset names (include SHA or build id in filename) so assets are immutable and the index always points to a specific one. Old assets can be pruned after deploy.
- Client retry: on `mip:digestMismatch`, re-fetch the index with cache-busting and retry once before erroring.
- Staging → promote: upload to a staging tag, update index, then move to the public tag.

magland:
@danfortunato

This is a serious issue.

Those possible fixes were suggested by claude, and I don't think I like any of them.

We'll need to think about this.

magland:
Dan and I decided that for now we will disable the sha-256 verification check. We should comment it out in the mip source.

PR #241, opened 2026-05-01
Require HTTPS for .mhl downloads and --url installs

### LINKED ISSUE #229: mhl_url from channel index should be required to be HTTPS
`+mip/+channel/download_mhl.m:33` accepts both schemes:

```matlab
isURL = startsWith(source, {'http://', 'https://'});
```

`mhl_url` comes from a channel's `index.json`. Anyone can host a third-party channel under `--channel owner/channel`, and nothing in mip rejects an `http://` value. Combined with SHA-256 verification being disabled (#201), a network attacker can swap `.mhl` payloads end-to-end. Loading the package then adds MEX/.m files to the path, so this is persistent code execution.

Proposed fix: in `download_mhl` and `installFromUrlFlag`'s `isZipUrl`, require `https://`.

magland:
Closes #229.

`mhl_url` comes from a channel's `index.json`, and any third-party channel can be added via `--channel owner/channel`. Combined with SHA-256 verification being disabled (#201), accepting `http://` lets a network attacker swap the `.mhl` payload end-to-end. Loading the package adds MEX/.m files to the path — persistent code execution.

## Changes

- `+mip/+channel/download_mhl.m` — reject `http://` with `mip:downloadMhl:requireHttps`
- `+mip/install.m` `installFromUrlFlag` — reject `http://` with `mip:install:requireHttps`
- `isZipUrl` — tightened to `https://` only as defense-in-depth
- Tests added for both rejection paths

### DIFF
--- a/+mip/+channel/download_mhl.m
+++ b/+mip/+channel/download_mhl.m
@@ -2,7 +2,10 @@ function localPath = download_mhl(source, destDir, expectedSha256)
 %DOWNLOAD_MHL   Download a .mhl package file from URL or copy from local path.
 %
 % Args:
-%   source          - URL (http:// or https://) or local file path
+%   source          - HTTPS URL or local file path. Plain http:// is
+%                     rejected (a network attacker could otherwise swap
+%                     .mhl payloads, leading to arbitrary code execution
+%                     once the package is loaded).
 %   destDir         - Destination directory for the downloaded file
 %   expectedSha256  - (Optional) Expected hex SHA-256 of the file (case-insensitive).
 %                     If provided and nonempty, the downloaded/copied file is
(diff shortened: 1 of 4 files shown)


magland:

The fix correctly addresses both download surfaces: `download_mhl` (used when installing from a channel's `index.json`) and `installFromUrlFlag` (used for `--url` installs). In both cases the `http://` check fires before any network activity.

The `installFromUrlFlag` check comes before FEX URL resolution, so `http://www.mathworks.com/fileexchange/...` input gets the https error immediately. After FEX resolution, `isZipUrl` (now tightened to `https://` only) provides a secondary check on the resolved URL.

`+mip/+channel/fetch_index.m` always constructs `https://` URLs via `mip.index()`, so the index-fetch path is unaffected.

One dead-code note (not a correctness issue): `isFexUrl` still recognizes `http://www.mathworks.com/...` as a valid FEX URL, but that branch is unreachable since the `requireHttps` check in `installFromUrlFlag` fires first.

No serious problems found.

PR #242, opened 2026-05-01
Make pin block all mip update paths (#228)

### LINKED ISSUE #228: mip update <pkg> ignores pin; design question for non-numeric pins
`mip pin` only blocks `mip update --all` (`+mip/update.m:78-88`). A direct `mip update foo` (and `mip update --deps foo`, and `mip update mip`) bypasses the pin entirely and updates the package in place. The only place `is_pinned` is consulted on the named-update path is to *unpin* after a `--force` update completes (`update.m:235`, `update.m:257`).

Docstring on `pin.m` is honest about this ("Pinned packages are skipped by `mip update --all`"), but it's a weaker guarantee than most users will expect from "pin."

Open question for v1: what should `mip update foo` do when `foo` is pinned?

For numeric pins (e.g. pinned at `1.2.0`) the natural behavior is to refuse and tell the user to `unpin` or pass `--force` — matches `--all` semantics.

For non-numeric pins (branch refs like `main`, `dev`) it's less clear. The pinned ref can drift on the channel side, and a user pinning to `main` plausibly *wants* `mip update foo` to fetch the latest commit on that ref. Two options:

1. **Strict**: pin always blocks update; user must unpin or `--force` regardless of ref type. Simple, predictable. Users who want "track this branch" use a different mechanism.
2. **Tiered**: numeric pin blocks; non-numeric pin allows update-in-place to the latest commit/tag on the same ref, but won't move to a different version.

Either way, the named-update path should consult `is_pinned`. Filing to decide on the semantics before v1.

magland:
Closes #228.

A pinned package is now refused by every `mip update` path. To update a pinned package, the user must `mip unpin <pkg>` first; `--force` does not override.

## Behavior

- `mip update <pkg>` (with or without `--force`): if `<pkg>` is pinned, raise `mip:update:pinned` before any other work, with a message instructing the user to run `mip unpin <pkg>` first.
- `mip update --all` (with or without `--force`): pinned packages are dropped from the batch with a "Skipping pinned package" message.
- `mip update --deps <pkg>`: if `<pkg>` itself is pinned, error as above. Pinned dependencies surfaced by the `--deps` expansion are dropped with a "Skipping pinned dependency" message.
- `mip update mip` follows the explicit-name rule (errors if `mip` is pinned).
- The previous "`--force` auto-unpins after a successful update" behavior is removed.

## Files

- `+mip/update.m`: enforce the pin rule on the named-update path; always filter pinned in `--all`; drop pinned deps from `--deps` expansion; remove dead auto-unpin code.
- `+mip/pin.m`, `+mip/unpin.m`: docstrings reflect strict block semantics.
- `docs/specification.md`: §7.1, §7.11.3, and the persistent-state file table updated.
- `tests/TestPinPackage.m`: 8 new tests covering the new error path and the `--deps` / `--all --force` skip behavior.

### DIFF
--- a/+mip/pin.m
+++ b/+mip/pin.m
@@ -6,8 +6,10 @@ function pin(varargin)
 %   mip.pin('org/channel/packageName')
 %   mip.pin('package1', 'package2')
 %
-% Pinned packages are skipped by "mip update --all". Use "mip update
-% --force" to override the pin (which also unpins the package).
+% Pinned packages are blocked from being updated by any "mip update"
+% invocation, including "mip update <pkg>", "mip update --force <pkg>",
+% "mip update --deps", and "mip update --all". To update a pinned
+% package, run "mip unpin <pkg>" first.
 %
 % Accepts both bare package names and fully qualified names.
 
--- a/+mip/unpin.m
+++ b/+mip/unpin.m
@@ -6,7 +6,7 @@ function unpin(varargin)
 %   mip.unpin('org/channel/packageName')
 %   mip.unpin('package1', 'package2')
 %
-% Unpinned packages will be updated normally by "mip update --all".
+% Unpinned packages will be updated normally by "mip update".
 %
 % Accepts both bare package names and fully qualified names.
 
--- a/+mip/update.m
+++ b/+mip/update.m
@@ -90,12 +90,21 @@ function update(varargin)
         end
         args = filtered;
     else
-        % Explicitly named packages: error immediately if any are
-        % pinned. The user must "mip unpin <pkg>" first; --force does
-        % not override the pin.
+        % Explicitly named packages: skip pinned packages with a
+        % "Skipping" message rather than erroring on the whole batch,
+        % so that "mip update X Y Z" with Y pinned still updates X and
+        % Z. The user must "mip unpin Y" first to update it; --force
+        % does not override the pin.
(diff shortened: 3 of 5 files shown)


magland:

The change consistently enforces "pin blocks all update paths" across the four entry points (`mip update <pkg>`, `--force`, `--all [--force]`, `--deps`) and removes the dead auto-unpin code. The logic flow is right:

- Explicit-name path raises `mip:update:pinned` before any work via `errorIfPinned`. Bare names, FQNs, and `--force` all go through this gate.
- `--all` always filters pinned (no longer gated on `~force`).
- `--deps` runs `errorIfPinned` on the explicit args first, then `expandWithDeps` drops pinned transitive deps with a message — so a user updating one package can't be silently blocked by an unrelated pinned dep.
- `mip update mip` is treated as an explicit named update, so the pin check happens before `updateSelf` is reached at line 145.

`is_pinned` canonicalizes its argument and `resolve_to_installed` returns the canonical FQN, so case/shorthand variants resolve correctly.

Test coverage is solid: pinned-named errors with and without `--force`, batch errors atomically when any one is pinned, error message mentions `unpin`, `--all --force` still skips, `--deps` drops pinned deps, `--deps` on a pinned explicit errors.

No serious problems found.

PR #243, opened 2026-05-01
Fix compare_versions for non-numeric versions

### LINKED ISSUE #227: compare_versions returns 0 for any non-numeric version, breaking info sort
`+mip/+resolve/compare_versions.m:14-15` parses both versions with `str2double`. For non-numeric strings (e.g. `'main'`, `'dev'`), `str2double` returns `NaN`, and all comparisons with `NaN` are false — so the function returns `0` (equal) for any pair involving a non-numeric version.

Impact: `+mip/info.m:319` (`sortVersions`) is an insertion sort keyed on `compare_versions(...) > 0`. Since that's never true when a non-numeric version is involved, mixed-version lists come out in whatever order `keys(versionMap)` produced — i.e. ASCII alphabetical — and `allVersions{end}` is used as the "latest" for the displayed dependency list.

Confirmed in MATLAB:

```
>> str2double('main')
ans = NaN
>> mip.resolve.compare_versions('2.0', 'main')
ans = 0
```

With versions `{'0.5','1.0','2.0','main'}`, `sortVersions` returns the input unchanged and `latest = 'main'` — so `mip info <pkg>` displays the dependencies of the `main` variant rather than the latest numeric release.

`select_best_version` is unaffected because it pre-filters to numeric versions before sorting.

Suggested fix: in `compare_versions`, detect `NaN` components and rank non-numeric strings below any numeric version (matching the tier ordering already used by `select_best_version`).

magland:
## Summary

`compare_versions` parsed both inputs with `str2double` and compared component-wise. For non-numeric strings (`'main'`, `'dev'`, ...) the components are `NaN`, all comparisons with `NaN` are false, so the function returned `0` (equal) for any pair involving a non-numeric version.

`info.m`'s `sortVersions` is keyed on `compare_versions(...) > 0`, so mixed-version lists came out in `keys(versionMap)` order (ASCII alphabetical) and `mip info <pkg>` displayed dependencies of e.g. `main` as the latest variant.

Fix: classify with `is_numeric_version` and define explicit tiers — numeric > `main` > `master` > other named (alphabetically first ranks higher). Numeric vs numeric comparison is unchanged.

Closes #227


magland:

The diagnosis and core fix are correct: the previous version's component-wise `str2double` comparison silently returned `0` for any non-numeric input, so `info.m`'s `sortVersions` was effectively a no-op for mixed-version lists.

One thing worth flagging: the new tier ordering diverges from `select_best_version`. `select_best_version` has an explicit `'unspecified'` tier (above "alphabetically first"); the new `compare_versions` does not. So for inputs like `{'unspecified', 'dev'}`:

- `select_best_version` picks `unspecified` (it's a named tier).
- `compare_versions`-based sorting picks `dev` as highest (both are tier 3, alphabetical-first ranks higher).

The test `testUnspecifiedIsOrdinaryName` documents this as intentional. If that's deliberate, `select_best_version` should arguably be aligned to match (drop the special `unspecified` tier). If it's not, `compare_versions` should treat `unspecified` as its own tier between `master` and other names. Either way, the two functions should agree on ordering, since both feed "what's the latest version" decisions in different code paths.

Otherwise the change looks good. Tests cover the meaningful cases (numeric component ordering, mixed tiers, the issue #227 sortVersions scenario).

PR #244, opened 2026-05-01
Fix infinite recursion in find_all_dependencies on cycles

### LINKED ISSUE #226: find_all_dependencies infinite-recurses on circular dependencies
`+mip/+dependency/find_all_dependencies.m:42` checks `~ismember(depFqn, deps)` against the *local* `deps` list, which starts empty in every recursive call. A cycle A → B → A re-enters with no memory of A and recurses until MATLAB's stack is exhausted.

Callers affected:
- `+mip/+state/prune_unused_packages.m:27` (runs during `uninstall` and `update`)
- `+mip/update.m:662` (`update --deps`)
- `+mip/unload.m:237`

`build_dependency_graph` already handles cycles via a threaded `visited`/`path` argument. `find_all_dependencies` should adopt the same pattern.

Repro: install two local packages whose `mip.json` files list each other as dependencies, then run `mip uninstall` on either.

magland:
## Summary
- `find_all_dependencies` checked `~ismember(depFqn, deps)` against the *local* `deps` list, which restarts empty in each recursive call. A cycle A → B → A re-entered with no memory of A and recursed until MATLAB's stack was exhausted.
- Threaded a `visited` set through recursion (same pattern as `build_dependency_graph`) so cycles terminate.
- Added `TestFindAllDependencies` covering no-deps, transitive, diamond, self-loop, direct, and indirect cycles.

Closes #226


magland:

Diagnosis is right: the previous `~ismember(depFqn, deps)` check operated on the *local* `deps` cell array, which is freshly empty in each recursive frame, so no cycle could ever be detected. Threading a `visited` set through recursion (mirroring the pattern already used in `build_dependency_graph`) is the correct fix.

The recursive call passes `[visited, deps]` rather than just `visited`, which is fine — it's a minor optimization that prevents re-traversing siblings whose subtrees have already been enumerated at the current node, and the outer-caller `ismember(depFqn, deps)` check still guards correctness.

Tests cover the meaningful cases: no deps, transitive, direct cycle (A→B→A), self cycle (A→A), indirect cycle (A→B→C→A), and diamond. The diamond test verifies the optimization above doesn't drop shared deps.

No serious problems found.

PR #245, opened 2026-05-01
Mark directly-loaded packages as directly_installed (#224)

magland:
Closes the second half of #224 (the first half was fixed in #225).

When a user runs `mip load <pkg>` on a package that was previously pulled in only as a transitive dependency, the package is now also marked as `directly_installed`. A later `mip uninstall` of the parent will no longer prune it.

Per @danfortunato's final comment on the issue: "Jeremy and I decided that `mip load` will mark the package as directly installed."

Both load paths in `loadSingle` (already-loaded promotion + fresh load) gain a `mip.state.add_directly_installed(fqn)` call gated on `isDirect`, so transitive dependency loads (`--transitive`) are unaffected.

### DIFF
--- a/+mip/load.m
+++ b/+mip/load.m
@@ -192,6 +192,9 @@ function matched = loadSingle(packageArg, installIfMissing, stickyPackage, chann
         else
             fprintf('Package "%s" is already loaded\n', displayFqn);
         end
+        if isDirect
+            mip.state.add_directly_installed(fqn);
+        end
         % If --sticky was specified, add to sticky packages
         if stickyPackage
             if ~mip.state.is_sticky(fqn)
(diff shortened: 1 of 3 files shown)


magland:

The change is minimal and well-targeted: both load paths in `loadSingle` (the already-loaded promotion path and the fresh-load path) now call `mip.state.add_directly_installed(fqn)` when `isDirect` is true. The internal `--transitive` path passes `isDirect=false` for recursive dependency loads, so transitive loads correctly do not promote.

`add_directly_installed` is idempotent (checks `ismember` before writing), so the unconditional call when `isDirect` is true is safe even when the package is already a root.

The four new tests cover the meaningful scenarios: direct load marks, transitive load does not mark, transitive-then-direct promotes, and the full issue #224 scenario where a directly-loaded transitive dep survives uninstall of its parent.

No serious problems found.

PR #256, opened 2026-05-02
Print per-package output blocks for mip update and mip uninstall

### LINKED ISSUE #248: Restructure mip update and mip uninstall to print per-package output blocks
`mip.update` and `mip.uninstall` both split their per-package output across multiple batched phases, so the output for one package is interleaved with output for the next. This is inconsistent with `mip.install` and `mip.load` (and `mip.unload`, `mip.pin`, `mip.unpin`), which already iterate per-package after any unavoidable up-front planning step.

### `mip.update`

Two batched phases:

1. For every package in the batch, fetch the channel index and decide whether it needs updating, printing `Checking for updates to "X" ...` / `Updating "X": a -> b` for each.
2. Then for every package that needs updating, download and replace, printing `Downloading X ...` / `Successfully updated "X"`.

For `mip update X Z` (both updating), output is:

```
Checking for updates to "X" ...
Updating "X": a -> b
Checking for updates to "Z" ...
Updating "Z": a -> b
Downloading X ...
Successfully updated "X"
Downloading Z ...
Successfully updated "Z"
```

X's check and download lines are split by Z's check. With pinning skip messages mixed in (see #228 / #242), this gets harder to read.

### `mip.uninstall`

Same pattern: unload all first, then uninstall all. For `mip uninstall X Y`:

```
Unloaded package "X"
Unloaded package "Y"
Uninstalling "X"...
Uninstalled package "X"
Uninstalling "Y"...
Uninstalled package "Y"
```

### Proposal

Bring both commands in line with `mip.install` / `mip.load`: collapse the per-batch loops into a single per-package loop that walks the batch in argument order and runs each package's full lifecycle before moving to the next. Whole-batch operations (`installMissingDeps`, `prune_unused_packages`, `check_broken_dependencies`) stay batched at the end of the loop where they belong.

Target output for `mip update X Y Z W` with Y, W pinned and X, Z both updating:

```
Checking for updates to "X" ...
Updating "X": a -> b
Downloading X ...
Successfully updated "X" to b
Skipping pinned package "Y". Run "mip unpin Y" first to allow updates.
Checking for updates to "Z" ...
Updating "Z": a -> b
Downloading Z ...
Successfully updated "Z" to b
Skipping pinned package "W". Run "mip unpin W" first to allow updates.
```

Target output for `mip uninstall X Y`:

```
Unloaded package "X"
Uninstalling "X"...
Uninstalled package "X"
Unloaded package "Y"
Uninstalling "Y"...
Uninstalled package "Y"
```

### Trade-off for `mip.update`

Today, `checkRemoteNeedsUpdate` runs for every remote package before any download begins, so a `mip:update:notInIndex` or `mip:update:versionNotInChannel` failure short-circuits the whole batch with nothing modified. After the restructure, an earlier package may be downloaded and replaced before a later package's check fails. The existing `try/catch` + `reloadPreviouslyLoaded` safety net handles partial failure (successfully-updated packages get reloaded, then the original error is re-raised), so this is a behavior change in *when* the failure surfaces, not in safety. Update the spec (§7.1) accordingly.

There is no analogous trade-off for `mip.uninstall` — its unload phase already calls `mip.unload(fqn)` on each package, and switching the order does not affect failure semantics.

### Files likely to change

- `+mip/update.m` — collapse the four current loops (self-update extraction, needs-update check, local update, remote download) into one per-package walk; extract a `updateLocalPackage` helper to keep the main loop readable.
- `+mip/uninstall.m` — merge the unload loop and the uninstall loop into one.
- `tests/TestPinPackage.m` — add an ordering test that pins some packages and asserts via `strfind` index comparison that skip lines appear at their argument-order slot.
- `tests/TestUninstallPackage.m` — add an ordering test (if not already covered) for multi-package uninstall.
- `docs/specification.md` — §7.1 step 7 (update phases) and the relevant uninstall section.

danfortunato:
## Summary

- Walk the batch in argument order for both commands so each package's full lifecycle prints as a contiguous block instead of interleaving across batched phases.
- `mip update`: pre-pass `classifyArg` keeps validation errors (`notInstalled`, `sourceNotFound`, `noCompileRequiresEditable`) raising before any destructive action; per-package loop now interleaves pin-skip / no-source-skip / "Checking for updates" / download lines in argument order. `installMissingDeps` and `prune_unused_packages` stay batched at end-of-loop -- they are decisions about the post-batch graph state.
- `mip uninstall`: merged the unload loop and uninstall loop into one per-package walk.

### Trade-off

A `mip:update:notInIndex` / `mip:update:versionNotInChannel` failure raised on a later package in the batch no longer short-circuits the whole batch -- earlier packages may already have been replaced on disk. The existing `try/catch` + `reloadPreviouslyLoaded` safety net still guarantees successfully-updated packages are reloaded before the original error is re-raised. Spec §7.1 updated to describe this.

Closes #248

## Test plan

- [x] `run_tests()` passes (671/671).
- [x] New `testUpdateNamed_PreservesArgumentOrderInterleaved` in `TestPinPackage.m` -- pins beta and delta, runs `mip.update('local/alpha', 'local/beta', 'local/gamma', 'local/delta')`, asserts each package's skip line appears in argument order.
- [x] New `testUninstall_PerPackageOrderInterleaving` in `TestUninstallPackage.m` -- loads alpha + beta, runs `mip.uninstall(...)`, asserts alpha's full Unloaded/Uninstalled lifecycle precedes beta's Unloaded line.

### DIFF
--- a/+mip/uninstall.m
+++ b/+mip/uninstall.m
@@ -80,26 +80,25 @@ function uninstall(varargin)
         return
     end
 
-    % Unload any packages that are currently loaded
-    for i = 1:length(resolvedPackages)
-        fqn = resolvedPackages{i};
-        if mip.state.is_loaded(fqn)
-            mip.unload(fqn);
-        end
-    end
-
-    % Uninstall each requested package
+    % Run each package's full lifecycle (unload, then uninstall) in
+    % argument order so the output for one package is not interleaved
+    % with the next.
     for i = 1:length(resolvedPackages)
         fqn = resolvedPackages{i};
         pkgDir = mip.paths.get_package_dir(fqn);
+        displayFqn = mip.parse.display_fqn(fqn);
+
+        if mip.state.is_loaded(fqn)
+            mip.unload(fqn);
+        end
 
         try
-            fprintf('Uninstalling "%s"...\n', mip.parse.display_fqn(fqn));
+            fprintf('Uninstalling "%s"...\n', displayFqn);
             rmdir(pkgDir, 's');
-            fprintf('Uninstalled package "%s"\n', mip.parse.display_fqn(fqn));
+            fprintf('Uninstalled package "%s"\n', displayFqn);
         catch ME
             error('mip:uninstallFailed', ...
-                  'Failed to uninstall package "%s": %s', mip.parse.display_fqn(fqn), ME.message);
+                  'Failed to uninstall package "%s": %s', displayFqn, ME.message);
         end
 
         % Remove from directly installed and pinned packages
--- a/+mip/update.m
+++ b/+mip/update.m
@@ -63,7 +63,12 @@ function update(varargin)
         end
     end
 
-    % --all: update all installed packages
+    % --all: expand to all installed packages. Pinned packages are
+    % filtered up-front for --all (the user did not specify an explicit
+    % order, so there is no per-arg position to anchor the skip messages
+    % to). Explicitly named pinned packages are handled inside the main
+    % per-package loop below so their skip messages appear in argument
+    % order, interleaved with the unpinned packages' update output.
     if updateAll
         if ~isempty(args)
             error('mip:update:allWithPackages', ...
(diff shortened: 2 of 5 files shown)


magland:
LGTM. The refactor (classifyArg + updateLocalPackage) is a clear improvement and the new tests pin down the argument-order contract. Two things worth noting post-merge: (1) consider adding a test that verifies earlier packages stay updated when a later one errors with versionNotInChannel — locks in the new partial-batch contract; (2) `mip update --no-compile mip` no longer errors and silently ignores the flag, since --no-compile validation now only runs over 'process' items, not 'self-update'. Probably fine but flagging.

Issue #250, opened 2026-05-01
`--channel <single_name>` should parse to the Github repo `single_name/mip-single_name`, not `mip-org/mip-single_name`


danfortunato:
Currently, `--channel foo` is parsed to the channel repo `mip-org/mip-foo`. Instead, we should parse this as shorthand for a Github user's personal mip channel `github.com/foo/mip-foo`.

Note that this shorthand only applies to `--channel` and not to FQNs, i.e. `foo/package1` is not shorthand for `foo/foo/package1`.

PR #257, opened 2026-05-02
Add --channel <name> shorthand for user's personal channel (closes #250)

danfortunato:
## Summary
- When `--channel` is given a single bare name (no `/`), expand it to `<name>/<name>` as shorthand for the user's personal channel repo at `github.com/<name>/mip-<name>`.
- The shorthand is exclusive to the `--channel` flag — FQN package args like `foo/pkg1` are unaffected.
- Implemented in [+mip/+parse/parse_channel_flag.m](+mip/+parse/parse_channel_flag.m) and the parallel `--channel` parser in [+mip/load.m](+mip/load.m).

## Test plan
- [x] New `testParseChannelFlagBareNameExpands` verifies `--channel foo` expands to `foo/foo`.
- [x] New `testParseChannelShorthandDoesNotApplyToFqn` verifies the shorthand doesn't leak into FQN parsing (`foo/pkg1` stays a 2-part non-gh FQN).
- [x] Updated existing `testParseChannelFlagPresent` and `testParseChannelFlagAtEnd` to assert the new expansion.
- [x] Full test suite passes (671/671).


danfortunato:
I wonder if the default channel for `--channel <name>` should be `<name>/core` instead of `<name>/<name>`. Then mip-org/core is automatically the default for mip-org. @magland what do you think?

magland:
> I wonder if the default channel for `--channel <name>` should be `<name>/core` instead of `<name>/<name>`. Then mip-org/core is automatically the default for mip-org. @magland what do you think?

I feel like "core" is the singular default channel that does not require specifying --channel. I don't think of `--channel user` having that same singular sense. It's just the default channel for that user. It's not a "core" channel. So I'd be in favor of `owner/owner` or `owner/default`. Both of those have the flavor to me of being the default channel for that owner.

danfortunato:
Yeah makes sense. Let's keep `<owner>/<owner>` then.

magland:
LGTM. Naming question is settled and the implementation matches. The shorthand-doesn't-leak-into-FQN-parsing test is a nice guard.

PR #261, opened 2026-05-05
mip test: prefer loaded package over installed-but-unloaded

### LINKED ISSUE #260: mip test X should use the loaded X with priority over an installed X that is not loaded
For example

mip load --channel mip-org/staging chebfun
mip test chebfun

Right now, that loads and tests the installed chebfun from core.

Also, if there are multiple loaded packages with that bare name, it should use the most recently loaded one for the test.

magland:
## Summary

Closes #260.

`mip test <X>` previously resolved bare names via the installed-package priority (gh/mip-org/core first, then alphabetical), so

```
mip load --channel mip-org/staging chebfun
mip test chebfun
```

would test the unrelated installed `mip-org/core/chebfun` instead of the chebfun the user just loaded.

Bare-name resolution for `mip test` now:

1. Searches `MIP_LOADED_PACKAGES` first; among multiple loaded matches, the **most recently loaded** wins.
2. Falls back to `resolve_bare_name` (gh/mip-org/core first, then alphabetical) when nothing matching is loaded.

FQN args (`mip test mip-org/core/chebfun`) bypass the loaded-priority logic and resolve directly against installed packages.

Spec: new §2.4.6 in the resolution-context catalog; §9.7 cross-references it.

### DIFF
--- a/+mip/test.m
+++ b/+mip/test.m
@@ -11,7 +11,11 @@ function test(varargin)
 %
 % The test script should error on failure and print 'SUCCESS' on success.
 %
-% Accepts both bare package names and fully qualified names.
+% Accepts both bare package names and fully qualified names. Bare-name
+% resolution prefers a currently loaded package over an installed-but-
+% unloaded one; if multiple loaded packages share the bare name, the most
+% recently loaded one wins. Otherwise, falls back to the installed-package
+% priority (gh/mip-org/core first, then alphabetical).
 
 if nargin < 1
     error('mip:test:noPackage', 'Package name is required for test command.');
@@ -19,8 +23,7 @@ end
 
 packageArg = varargin{1};
 
-% Resolve to installed FQN
-r = mip.resolve.resolve_to_installed(packageArg);
+r = resolveTestTarget(packageArg);
 if isempty(r)
     error('mip:test:notInstalled', ...
           'Package "%s" is not installed.', packageArg);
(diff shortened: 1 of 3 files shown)


magland:
Review by Claude (Opus 4.7).

The change is small, well-scoped, and the logic looks correct:

- `resolveTestTarget` correctly short-circuits on FQN args, walks `MIP_LOADED_PACKAGES` with `mip.name.match` (which respects the case- and dash/underscore-insensitivity rule), picks the last match, and falls back to `resolve_to_installed` otherwise.
- The fallback is preserved exactly — if nothing is loaded, behavior is unchanged.
- The four new tests cover the meaningful cases: loaded-wins, most-recent-among-multiple, no-loaded fallback to core, and FQN bypass. The FQN bypass test is the one I most wanted to see, and it's there.
- Spec §2.4.6 and §9.7 are updated and cross-linked.

No genuine issues found. A couple of edge cases I considered and don't think are worth blocking on:

- "Most recently loaded" means "last appended to `MIP_LOADED_PACKAGES`", not "last activated by the user". A re-`mip load` of an already-loaded package is a no-op and won't move it to the end. This matches the spec wording ("last in the `MIP_LOADED_PACKAGES` list") and seems fine.
- If a package is in `MIP_LOADED_PACKAGES` but its directory has gone missing, `resolve_to_installed` will return empty and `mip test` errors out instead of falling through to the bare-name fallback. That's a broken-state scenario and erroring seems reasonable.

LGTM.

PR #262, opened 2026-05-06
Add channel subscriptions for bare-name install resolution

### LINKED ISSUE #259: Add channel subscriptions for bare-name resolution
Add the ability to subscribe to channels so bare-name installs can fall back beyond `mip-org/core`.

New commands:
- `mip subscribe <channel>` — subscribe to a channel (`org/channel` or `<name>` shorthand for `<name>/<name>`)
- `mip unsubscribe <channel>` — unsubscribe
- `mip subscriptions` — list subscribed channels in subscription order

Behavior change: when `mip install <bare-name>` is invoked without `--channel`, resolution checks `mip-org/core` first, then each subscribed channel in the order they were subscribed. The first channel that publishes the package wins. Explicit `--channel` and FQNs are unaffected.

Subscriptions persist across sessions (e.g. a `subscriptions.txt` file under the mip root, similar to `directly_installed.txt`).

magland:
Closes #259.

## Summary

- New commands: `mip channel add <channel>`, `mip channel remove <channel>`, `mip channel list`.
- When `mip install <bare-name>` is invoked without `--channel`, resolution consults `mip-org/core` first, then each subscribed channel in priority order. The first channel that publishes the package wins. Explicit `--channel` and FQNs are unaffected.
- Re-subscribing an already-subscribed channel moves it to the top of the priority list (per discussion in #259).
- Subscriptions persist at `<root>/packages/channels.txt`. The `<name>` shorthand for `<name>/<name>` is honored.
- Bare-name dependencies of installed packages continue to resolve to `mip-org/core` — only top-level user args walk the priority list.


magland:
Review by Claude (Opus 4.7). A few real / potential issues — leaving as one comment per request.

**1. `mip help channel` is broken.** [mip.m:155-162](mip.m#L155-L162) dispatches `help` by looking for `+mip/<cmd>.m`. Every other documented command has such a file; `channel` does not (and as the comment in `handle_channel_command` notes, the `+mip/+channel/` subpackage prevents creating one). Result: `mip help channel` errors with `mip:unknownCommand`, even though `channel` is now listed in the top-level usage. Either special-case `'channel'` in the help dispatch or stash a help docstring somewhere the dispatcher can find (e.g., `+mip/channel_help.m`, or print inline from `handle_channel_command`).

**2. `get_channels` silently returns `{}` when `fopen` fails on an existing file.** [+mip/+state/get_channels.m:24-27](+mip/+state/get_channels.m#L24-L27) — the `~exist(...)` branch is fine, but if the file *does* exist and `fopen` returns -1 (permissions, transient lock, etc.), the function silently returns empty. The next `add_channel` call will then write a one-entry `channels.txt` via `set_channels`, which atomically replaces the original — silently dropping the user's prior subscriptions. Recommend erroring (or at minimum warning loudly) when `exist(...)` is true but `fopen` fails.

**3. `normalizeChannel` is duplicated verbatim** between [+mip/+state/add_channel.m:29-43](+mip/+state/add_channel.m#L29-L43) and [+mip/+state/remove_channel.m:26-40](+mip/+state/remove_channel.m#L26-L40). Validation and shorthand-expansion rules in two places will drift. Worth pulling into a shared helper (e.g. `+mip/+parse/normalize_channel_spec.m`, alongside `parse_channel_spec`).

**4. Channel indexes are fetched twice on the no-`--channel` path.** `resolveBareNameChannels` calls `mip.channel.fetch_index(chSpec)` for each priority-list channel ([+mip/install.m:705](+mip/install.m#L705)) but does not populate `packageInfoMap`; the later loop ([+mip/install.m:269-272](+mip/install.m#L269-L272)) calls `fetchChannelIndex` for the same channel and re-runs `fetch_index`. Currently absorbed by the 30 s disk cache in `fetch_index`, so it's not a network hit today — but it is a fragile coupling: a TTL change or cache-key change turns this into 2× HTTP per subscribed channel. Cleanest fix: have `resolveBareNameChannels` return the parsed indexes and feed them into `packageInfoMap` directly, rather than relying on the cache to dedupe.

Nothing else stood out — channel-flag override, FQN bypass, and the priority-order semantics all look right and the test coverage matches what the PR claims.

magland:
Follow-up by Claude (Opus 4.7). Pushed [31552c6](https://github.com/mip-org/mip/commit/31552c6504bb6715d5766086ad0c7e65a85db634) addressing items 1–3 from my prior review. Verified locally with `matlab -batch` against this branch — all 643 tests in `run_tests` still pass.

**1. `mip help channel`** — routed to a dedicated `print_channel_help()` since `+mip/+channel/` blocks the usual `+mip/<cmd>.m` help target. Confirmed `matlab -batch \"mip('help','channel')\"` now prints usage instead of erroring.

**2. `get_channels` on unreadable file** — now errors with `mip:fileError` when `channels.txt` exists but `fopen` returns -1. Verified the data-loss path: with two subscriptions present, `chmod 000` on `channels.txt`, then `mip channel add c/three` → previously silently rewrote the file to just `c/three`; now errors and the file is left untouched.

**3. Duplicated `normalizeChannel`** — extracted to `+mip/+parse/normalize_channel_spec.m`; `add_channel`/`remove_channel` both call it. `parse_channel_flag` left alone since its shorthand-expansion path doesn't validate `org/channel` shape and tightening it could affect unrelated callers — happy to follow up on that separately if you'd like.

**4. Double-fetch in `resolveBareNameChannels`** — left as-is. The cleanest fix is to share the parsed `chIndex`/populated maps between `resolveBareNameChannels` and `fetchChannelIndex`, which is a nontrivial refactor I didn't want to bundle into a review-followup commit. The 30 s `fetch_index` disk cache makes it a non-issue at runtime today.

magland:
Review by Claude (no nits, just real/potential issues).

A few things worth a look:

1. **Misleading "package not found" error after subscription walk.** In `resolveBareNameChannels`, an unresolved bare name is defaulted to `mip-org/core`, so the downstream error reads `Package "mip-org/core/foo" not found in repository`. A user who subscribed to `mylab/dev` and `otherco/extras` and ran `mip install foo` will be told the package wasn't in core — with no hint that the subscribed channels were also consulted. Worth surfacing the searched list (e.g. `not found in any of: mip-org/core, mylab/dev, otherco/extras`) so the priority-list behavior is observable when it fails.

2. **No locking on `channels.txt`.** `set_channels` does an atomic rename of a tmp file (good), but read-modify-write in `add_channel` / `remove_channel` is not protected. Two concurrent `mip channel add` invocations can lose updates (each reads the same baseline, each writes its own one-entry list). Probably rare for an interactive CLI, but worth acknowledging — at minimum a comment noting the assumed serial usage.

3. **`set_channels` doesn't check `fprintf` return value.** The try/catch handles thrown errors, but `fprintf` to a file generally returns a byte count rather than throwing on disk-full / quota-exceeded. A truncated tmp file would then be `movefile`'d into place, silently corrupting the subscriptions list. Checking the return of each `fprintf` (or a final `ftell`) would close that gap.

4. **Test coverage gap on the warning path.** `resolveBareNameChannels` warns and continues when a non-core subscribed channel's index fetch fails, but there's no test covering that branch — easy to regress into a hard error.

5. **Minor: `testChannels_PersistAcrossCalls` doesn't actually exercise persistence** — it calls `get_channels` twice in the same process and asserts equality. Reading the file fresh after dropping any in-memory state (or just checking the on-disk contents) would make it a real persistence test.

Nothing blocking; the design and tests are otherwise solid.

magland:
Pushed 585eaab addressing point 1 (the misleading "package not found" error).

`resolveBareNameChannels` no longer silently defaults unresolved bare names to `mip-org/core`. The caller now defaults them and tracks an `unresolvedBareIdx` mask, so the "not found" error path emits:

> `Package "foo" not found in any of: mip-org/core, mylab/dev, other/place`

instead of the previous `Package "mip-org/core/foo" not found in repository`. Behavior is unchanged when `--channel` is passed or for FQN args.

Added `testInstall_BareName_NotFound_ErrorListsConsultedChannels` to lock the new message in. All `TestChannelSubscriptions` and `TestInstall*` tests pass locally.

Leaving the other points from the earlier review for you to handle (or not) as you see fit.

magland:
Review by Claude. Two potential correctness concerns; otherwise the change looks clean and well-tested.

**1. Silent fallback when a subscribed channel's index fetch fails ([+mip/install.m](+mip/install.m) `resolveBareNameChannels`).** The `try/catch` warns and `continue`s for any non-core channel error. A transient fetch failure on a higher-priority channel can then silently resolve a bare name to a *lower*-priority channel that happens to publish the same name — installing a different package than the user intended. The warning is easy to miss in CI/scripts. Consider re-raising (or at minimum recording the failure and refusing to fall through to a same-named package in a lower-priority channel) so the user sees the error rather than getting a quietly-substituted install.

**2. Case-insensitive channel comparisons in `+mip/+state/add_channel.m` and `remove_channel.m` (`strcmpi`).** Per [+mip/+name/normalize.m](+mip/+name/normalize.m) and the project conventions in [CLAUDE.md](CLAUDE.md), owner/channel components are intentionally case-sensitive — they map to GitHub paths and to the on-disk `<root>/packages/gh/<owner>/<channel>/...` layout. Using `strcmpi` here means `mip channel add MyLab/dev` followed by `mip channel add mylab/dev` is treated as a re-subscribe (replacing the stored spelling), and the `mip-org/core` reject check would also collapse mixed-case variants. If a user has packages installed under one casing and subscribes under another, resolution will install into a sibling directory. Recommend `strcmp` for consistency with the rest of the codebase (and add a mixed-case test).

Minor (not blockers):
- `resolveBareNameChannels` only checks names; version constraints (`foo@1.5`) aren't considered when picking the channel. That's fine — the later install path raises a clear error — but worth a doc note in the channel-resolution comment.
- `get_channels`/`set_channels` round-trip stores whatever spelling the user typed; combined with #2 above, this can produce inconsistent state.

magland:
Pushed cdc817b addressing both real concerns from the review:

1. **`resolveBareNameChannels`**: removed the `try/catch warning/continue` for non-core channel fetch failures. Any subscribed-channel fetch error now aborts uniformly (same as core's existing behavior), so a transient failure on a higher-priority channel can no longer silently install a same-named package from a lower-priority one.
2. **`add_channel` / `remove_channel`**: switched the three `strcmpi` sites to `strcmp` to match the rest of the codebase's case-sensitive treatment of owner/channel components (per `mip.name.normalize` doc). Added `testChannels_AreCaseSensitive` to lock this in.

Broader case-insensitivity (e.g. for `--channel` and FQN args, which today preserve case and produce capitalized install paths) is left for a separate PR.

All 25 tests in `TestChannelSubscriptions` pass locally.

magland:
Reviewed the channel-subscription PR — design is clean and the test coverage is thorough. Two things worth a second look:

**1. A single unreachable subscribed channel aborts bare-name resolution even when another subscribed channel could resolve the name.**

Concrete scenario:

```
$ mip channel add mylab/dev      # subscribed first → lower priority
$ mip channel add other/place    # subscribed second → higher priority
# `foo` only exists in other/place. mylab/dev's GitHub Pages site is briefly down.
$ mip install foo
Fetching package index for mip-org/core...
Fetching package index for other/place...    # foo found here, bare name resolved
Fetching package index for mylab/dev...
Error: mip:indexFetchFailed — Failed to fetch package index from
  https://mylab.github.io/mip-dev/index.json: ...
```

That last fetch shouldn't even be needed — `foo` was already resolved by `other/place`. The cause is in `resolveBareNameChannels` (`+mip/install.m`): the early-exit check `anyUnresolved(...)` runs at the *top* of the loop, but the `fetch_index` call happens *before* we look at the index, so a broken channel still gets fetched in the iteration where we'd otherwise have broken out. Worse, even if `foo` weren't already resolved, an unreachable higher-priority subscription kills the install before consulting any lower-priority one — e.g.:

```
$ mip channel add other/place    # has foo
$ mip channel add mylab/dev      # most recent → highest priority, currently down
$ mip install foo
Error: mip:indexFetchFailed — ...mylab/dev...
# foo never gets a chance to resolve against other/place
```

Two policy options worth weighing:
- *Skip with warning:* `try`/`catch` around the per-channel `fetch_index` in `resolveBareNameChannels`; if it fails, log a warning and move on. If the name is ultimately unresolved, include the skipped channels in the not-found error.
- *Fail-fast (status quo):* keep current behavior, but at minimum re-order so the early-exit check runs before the fetch (saves the fetch in scenario 1 above).

**2. `add_channel`'s core-detection guard is case-sensitive — users can accidentally create a duplicate of the default channel.**

In `+mip/+state/add_channel.m:14`:

```matlab
if strcmp(channel, 'mip-org/core')
    fprintf('mip-org/core is the default channel and is always consulted first.\n');
    return
end
```

What happens with a case variant:

```
$ mip channel add MIP-ORG/CORE
Subscribed to channel "MIP-ORG/CORE".

$ mip channel list
Subscribed channels (in priority order):
  mip-org/core   (default, always first)
  MIP-ORG/CORE
```

Now `mip install <pkg>` will fetch GitHub Pages at the same upstream URL twice (Pages is case-insensitive on the host side), and any package resolved through the second entry installs to `<root>/packages/gh/MIP-ORG/CORE/...`, parallel to the canonical `gh/mip-org/core/...` tree. That breaks `mip list`/`mip load` consistency and silently doubles disk usage. Note this is distinct from the `MyLab/Dev` vs `mylab/dev` case in `testChannels_AreCaseSensitive` — those refer to two different user-controlled repos. `mip-org/core` is a known-singleton; the guard for *just that one channel* should match case-insensitively (`strcmpi`), e.g.:

```matlab
if strcmpi(channel, 'mip-org/core')
    ...
end
```

**Minor doc/code mismatch:** `+mip/+state/remove_channel.m`'s header comment says "No-op (with a warning) if the channel is not currently subscribed." The implementation prints with `fprintf`, not `warning(...)`:

```matlab
if ~any(mask)
    fprintf('Channel "%s" is not subscribed.\n', channel);  % header says "warning"
    return
end
```

Either change the call to `warning('mip:channelNotSubscribed', ...)` or update the docstring to say "informational message."

— review by Claude

magland:
Addressed review issue 1 and the doc nit in 64da674.

- **Issue 1 (skip-with-warning):** `resolveBareNameChannels` now wraps the per-channel `fetch_index` in `try`/`catch`. On failure it emits a `mip:channelUnreachable` warning naming the channel and the underlying error, records it in a new `skippedChannels` cell, and `continue`s. The "not found" error in `installFromRepository` appends `(could not fetch indexes for: <list>)` when any channels were skipped, so the user sees both what was searched and what was unavailable. The two scenarios from the review now behave as:

  ```
  $ mip channel add other/place    # has foo
  $ mip channel add mylab/dev      # most recent → highest priority, currently down
  $ mip install foo
  Warning: Could not fetch index for channel "mylab/dev" (Failed to fetch package
    index from https://mylab.github.io/mip-dev/index.json: ...). Skipping;
    bare-name resolution will continue with the remaining channels.
  Fetching package index for mip-org/core...
  Fetching package index for other/place...
  ...installs foo from other/place
  ```

- **Doc nit:** `remove_channel.m`'s header now says "prints an informational message" instead of "with a warning."

- **Issue 2 (case-sensitive core guard):** intentionally left as-is per follow-up direction.

Full test suite (711 tests) passes locally. I didn't add a new unit test for the unreachable-channel path because forcing a deterministic `fetch_index` failure in this test setup requires either real network access (flaky) or test-only injection that doesn't exist yet — happy to add one if you want to take that on as a follow-up.

magland:
Review by Claude (Opus 4.7).

Overall the PR looks solid — the priority-list semantics match the description, the test coverage is thorough (priority order, move-to-top, FQN/`--channel` overrides, mixed args, unreachable channels, error message contents), and I/O is appropriately atomic (tmpfile + `movefile -f`, `onCleanup` fclose).

A couple of things worth a look:

1. **Duplicated channel-index processing.** `resolveBareNameChannels` calls `mip.channel.fetch_index(chSpec)` and walks `chIndex.packages` to build `availableNorms`. Then for every channel a bare name resolves into, `fetchChannelIndex` is called from the main loop, which calls `mip.channel.fetch_index` again and runs `build_package_info_map` over the same index. The 30s on-disk cache in `fetch_index` keeps this off the network in practice, but the JSON parse + iteration runs twice per priority channel on every install. Threading the already-parsed `chIndex` through (or merging into `packageInfoMap` during the priority walk and reusing `fetchedChannels` so subsequent calls are no-ops) would avoid that.

2. **Inconsistent fetch-failure handling between phases.** During priority-list resolution (`resolveBareNameChannels`), a fetch failure warns and skips. After resolution, `fetchChannelIndex` is called for the resolved channel (e.g. core, or whichever subscribed channel the bare name landed in) and rethrows on failure. In normal operation the on-disk cache makes this fine, but if the cache TTL elapses between the two phases (or write-back to cache failed) you can get a soft warning during resolution and then a hard error fetching the same channel a moment later. Worth at least documenting; ideally the two paths share state.

3. **`normalize_channel_spec` doesn't reject whitespace-only parts.** `mip channel add ' / '` (or any all-whitespace owner/channel) passes the `cellfun('isempty', parts)` check and gets persisted verbatim. Probably want a `strtrim` + non-empty check on each part, or a stricter regex (`^[A-Za-z0-9._-]+$` per GitHub's allowed characters), so bad input fails at `add` time rather than later when the index fetch fails.

Minor: `add_channel` prints `Subscribed to channel "%s".` even when the channel was already subscribed and was just moved to the top of the priority list — slightly misleading for that path.

magland:
I think the remaining issues are too minor. @danfortunato

magland:
@danfortunato addressed both points from https://github.com/mip-org/mip/issues/259#issuecomment-4390290751 in 7d2c8eb:

- `mip channel list` always shows `mip-org/core` at the top as a uniform list entry.
- Added `mip channel append <channel>` — subscribes at lowest priority; re-running on an existing subscription moves it to the bottom (mirroring `add` → top).

PR #269, opened 2026-05-07
Store channels.txt in stack order (most recent last)

### LINKED ISSUE #268: Change the `channels.txt` to have most recently added at the end of the file
The user-facing print statements will show the most recently added channel at the top of the list, but the internal data structure in `channels.txt `should match the stack-like flavor of `MIP_LOADED_PACKAGES`.

magland:
Closes #268.

## Summary

- `channels.txt` now persists with the most recently added (highest-priority) channel at the END of the file, matching the append semantics of `MIP_LOADED_PACKAGES`.
- `set_channels` writes the in-memory cell array in reverse; `get_channels` reverses on read. Callers and the user-facing `mip channel list` output are unchanged.
- Added a test that asserts the on-disk file order directly.

Targets the `add-channel-subscriptions` branch since #262 (which introduces `channels.txt`) is still open.

### DIFF
--- a/+mip/+state/get_channels.m
+++ b/+mip/+state/get_channels.m
@@ -9,7 +9,10 @@ function channels = get_channels()
 %              need not be listed.
 %
 % Channels are persisted at <root>/packages/channels.txt, one channel
-% per line, in priority order.
+% per line, in stack-like order: the most recently added channel is at
+% the END of the file (matching the append semantics of
+% MIP_LOADED_PACKAGES). The returned cell array reverses that so the
+% highest-priority channel is at index 1.
 
     channels = {};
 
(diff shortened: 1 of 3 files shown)


magland:
Review by Claude (Opus 4.7, invoked by @magland).

Looks good — no correctness concerns.

- The encoding/decoding is symmetric: `set_channels` writes `length:-1:1` and `get_channels` `fliplr`s on read, so round-tripping a channel list is identity. Empty lists are also fine (`0:-1:1` is an empty range in MATLAB and `fliplr({})` is `{}`).
- I confirmed that `channels.txt` is only read/written via `get_channels`/`set_channels` (grep across the tree), so the on-disk format change is fully encapsulated.
- The new `testChannelsFile_StoresMostRecentLast` directly asserts the on-disk line order, which is the right place to lock this in. The existing `testAppendChannel_*` cases still pass under the new layout (traced by hand).
- Targeting the open `add-channel-subscriptions` base branch is the right call since `channels.txt` hasn't shipped yet — no migration concern for existing users.

PR #270, opened 2026-05-07
mip load: move already-loaded package to end of load order

### LINKED ISSUE #267: `mip load` on an already-loaded package should move it to the end of `MIP_LOADED_PACKAGES` (and `MIP_DIRECTLY_LOADED_PACKAGES`)

magland:
## Summary
- `mip load` on an already-loaded package now moves its FQN to the end of `MIP_LOADED_PACKAGES`.
- Direct re-loads also move it to the end of `MIP_DIRECTLY_LOADED_PACKAGES` (this subsumes the prior transitive→direct promotion path).
- New `mip.state.key_value_move_to_end` helper does remove + append so callers don't duplicate the logic.

Closes #267

### DIFF
--- a/+mip/+state/key_value_move_to_end.m
+++ b/+mip/+state/key_value_move_to_end.m
@@ -1,0 +1,12 @@
+function key_value_move_to_end(key, value)
+%KEY_VALUE_MOVE_TO_END   Move a value to the end of a key's list.
+%
+% If the value is not present, it is appended. If it is present, it is
+% removed and re-appended so it ends up at the end of the list.
+
+values = mip.state.key_value_get(key);
+values(ismember(values, value)) = [];
+values{end+1} = value;
+mip.state.key_value_set(key, values);
+
+end
--- a/+mip/load.m
+++ b/+mip/load.m
@@ -191,16 +191,23 @@ function matched = loadSingle(packageArg, installIfMissing, stickyPackage, chann
 
     % Check if package is already loaded
     if mip.state.is_loaded(fqn)
-        % If this is a direct load and the package was previously
-        % loaded as a dependency, mark it as direct now
-        if isDirect && ~mip.state.is_directly_loaded(fqn)
-            mip.state.key_value_append('MIP_DIRECTLY_LOADED_PACKAGES', fqn);
-            fprintf('Package "%s" is already loaded (now marked as direct)\n', displayFqn);
+        % Re-loading an already-loaded package moves it to the end of
+        % MIP_LOADED_PACKAGES so the load order reflects the most recent
+        % load. Direct re-loads do the same for MIP_DIRECTLY_LOADED_PACKAGES,
+        % which also covers the case where the package was previously loaded
+        % only as a transitive dependency and is now being promoted.
+        wasDirect = mip.state.is_directly_loaded(fqn);
+        mip.state.key_value_move_to_end('MIP_LOADED_PACKAGES', fqn);
+        if isDirect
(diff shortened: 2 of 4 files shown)


magland:
**Review by Claude (Opus 4.7) — automated review requested by @magland**

The change is small, well-scoped, and the core behavior matches the issue. Implementation of `key_value_move_to_end` mirrors the existing `key_value_remove`/`key_value_append` style and looks correct. One real issue worth addressing:

### Test does not exercise the path it names

`testLoadPackage_TransitiveReloadDoesNotMoveDirectlyLoaded` would pass against `main` without this PR's changes. The new `isDirect=false` branch of the `is_loaded` block in `loadSingle` is never reached in this test, because the deps loop in `+mip/load.m:255` short-circuits when a dep is already loaded:

```matlab
if ~mip.state.is_loaded(depFqn)
    loadSingle(depFqn, ..., false, ...);
else
    fprintf('  Dependency "%s" is already loaded\n', ...);
end
```

So when the test loads `mainpkg` with `depA` already loaded, `loadSingle` is never re-entered for `depA`, and the new transitive-reload branch (the `else` at `+mip/load.m:209-211`) is not executed. The properties the test verifies (`depA` still in `direct`, `mainpkg` at end) hold trivially regardless of the new code.

To actually cover the transitive re-load path, the test needs to force re-entry, e.g.:

```matlab
mip.load('--transitive', 'mip-org/core/depA');  % already loaded
```

and then verify that `MIP_LOADED_PACKAGES` was reordered (depA moved to end) while `MIP_DIRECTLY_LOADED_PACKAGES` was not.

### Minor observations (not blocking)

- The "promote transitive→direct" path now relies on `move_to_end` doing an append-when-absent. That's fine and matches the helper's documented behavior, but if `key_value_move_to_end` is ever changed to be a strict "must exist" operation, the promotion path will silently break. The behavior is locked in by `testMoveToEndAppendsWhenAbsent`, so this is just a thing to keep in mind.
- `mip.list` (`+mip/list.m:87-95`) uses `MIP_LOADED_PACKAGES` order for "most recent first" display — the new reordering makes that display more accurate, which is a nice side effect worth noting in the PR description if you want.

Otherwise LGTM.

magland:
Pushed [ff64e7c](https://github.com/mip-org/mip/commit/ff64e7c) — Claude addressing the test-coverage gap from the previous review.

Rewrote `testLoadPackage_TransitiveReloadDoesNotMoveDirectlyLoaded` to use `mip.load('mip-org/core/depA', '--transitive')`, which forces re-entry into `loadSingle` for an already-loaded package. The test now asserts both:

- `MIP_LOADED_PACKAGES` end is `depA` (moved on transitive re-load)
- `MIP_DIRECTLY_LOADED_PACKAGES` end is still `pkgB` (NOT moved, since the re-load was not direct)

Confirmed the previous form passed against `main` without the PR's changes; this form does not.

danfortunato:
Superseded — closing as redundant.

The two halves of this PR (issue #267) have both landed on `main` by other means:

- **`MIP_LOADED_PACKAGES` move-to-end on re-load** was independently implemented in c6126a2 ("Move already-loaded package to most-recent on re-load").
- **`MIP_DIRECTLY_LOADED_PACKAGES` move-to-end on a direct re-load** — the half `c6126a2` left out — is carried by #291, rebuilt cleanly on current `main` (without this PR's `key_value_move_to_end` helper, to keep the change minimal). #291 closes #267.

Thanks for the original work here — the test cases in #291 are derived from this branch.

PR #271, opened 2026-05-07
Atomically write functionSignatures.json (fixes #266)

### LINKED ISSUE #266: `mip` tab completion is sometimes broken
Typing
```
>> mip
```
can sometines show the error
```
Unable to load signatures for 'mip'. Failed to parse JSON file '/Users/dfortunato/Documents/MATLAB/mip/packages/gh/mip-org/core/mip/mip/resources/functionSignatures.json': <unspecified file>(1): expected value
```

magland:
## Summary

Fixes #266 — `mip` tab completion intermittently failed with:

```
Unable to load signatures for 'mip'. Failed to parse JSON file
'.../mip/resources/functionSignatures.json': <unspecified file>(1): expected value
```

`fopen(jsonPath, 'w')` truncates the destination file the moment it's opened, leaving a window where MATLAB's tab-completion reader can see an empty file before `fwrite` finishes. The "expected value" error at line 1 is exactly what an empty file produces.

## Changes

- `+mip/+state/update_function_signatures.m`: validate the generated JSON in memory with `jsondecode`, then write to a sibling `.tmp` file and `movefile` it into place. The destination is never truncated; readers either see the previous valid file or the new one.
- `tests/TestFunctionSignatures.m`: two regression tests — one verifies the generated file always parses as JSON, the other verifies repeated regeneration stays valid and leaves no `.tmp` debris.

### DIFF
--- a/+mip/+state/update_function_signatures.m
+++ b/+mip/+state/update_function_signatures.m
@@ -45,13 +45,38 @@ function update_function_signatures(targetResourcesDir)
 
     json = build_signatures(installed, loadable, loaded, pinned);
 
+    % Validate before writing: tab completion reads this file
+    % asynchronously, and a malformed file produces a "Failed to parse
+    % JSON" error in the Command Window.
+    try
+        jsondecode(json);
+    catch
+        return
+    end
+
+    % Write atomically. fopen(..., 'w') truncates the destination
+    % immediately, leaving a window where a concurrent reader sees an
+    % empty file. Stage to a sibling temp file and rename into place.
     jsonPath = fullfile(targetResourcesDir, 'functionSignatures.json');
-    fid = fopen(jsonPath, 'w');
+    tmpPath  = [jsonPath '.tmp'];
+
+    fid = fopen(tmpPath, 'w');
     if fid == -1
         return
     end
-    cleaner = onCleanup(@() fclose(fid));
-    fwrite(fid, json);
+    nWritten = fwrite(fid, json);
+    fclose(fid);
+    if nWritten ~= numel(json)
+        if exist(tmpPath, 'file')
+            delete(tmpPath);
+        end
+        return
+    end
+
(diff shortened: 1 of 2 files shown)


magland:
*Disclosure: I'm Claude (Anthropic AI assistant), doing this review at the author's request.*

The atomic-rename approach is the right shape for this race, and the regression tests directly cover #266. One thing worth checking before merging:

**`movefile` atomicity on non-POSIX platforms.** The reported failure is on macOS, where MATLAB's `movefile` is backed by `rename(2)` and gives you the property the PR is relying on — the reader sees either the old contents or the new contents, never an empty/partial file. On Windows the guarantee is weaker: depending on MATLAB version and whether the destination is open, `movefile` can fall back to a copy-then-delete path, which re-opens the same empty-file window the PR is trying to close. If supporting Windows here matters, `java.nio.file.Files.move(java.nio.file.Paths.get(tmpPath), java.nio.file.Paths.get(jsonPath), javaArray('java.nio.file.CopyOption', ...))` with `REPLACE_EXISTING` + `ATOMIC_MOVE` would throw on platforms that can't honor it, rather than silently degrade. If Windows isn't a concern in practice, this is fine as-is.

Two smaller observations, neither blocking:

- The `jsondecode(json)` pre-write check is independent of the concurrency fix — it guards against `build_signatures` ever producing malformed JSON (e.g., a package whose bare name contains a `"` would slip through `quote_one`, which only escapes MATLAB single quotes). The inline comment frames it as part of the async-reader story, which is a little misleading.
- Removing the `onCleanup(@() fclose(fid))` means a thrown `fwrite` would leak the fid, but `fwrite` essentially never throws after a successful `fopen`, so this is theoretical.

LGTM modulo the Windows question.

magland:
I'd say let's not worry about the windows potentially non-atomic movefile

PR #274, opened 2026-05-07
resolve_dependency: prefer depending package's own channel before core

### LINKED ISSUE #264: Bare-name dependencies should prefer the depending package's own channel before core
Recording a decision @danfortunato and I made earlier today.

Currently `mip.resolve.resolve_dependency` resolves any bare-name dep in `mip.yaml` directly to `gh/mip-org/core/<name>`. We decided that when resolving a dep for a package on `owner/chan`, it should check `owner/chan` first and fall back to `mip-org/core` if not found. Fully qualified deps continue to work as today.

We also decided to encourage bare names whenever possible. E.g. `flatironinstitute/chunkie` depends on `fmm2d` and `flam`, which currently live in `flatironinstitute` and will eventually move to `core`. With this change, `chunkie/mip.yaml` can list bare `fmm2d`/`flam`, and once those land in core no edit is needed — bare names automatically retarget.

Tradeoff we acknowledged: behavior of a package may shift when a dep moves to core (different version/build). Reviewers should manually load the core versions before approving the move.

Touches:
- `+mip/+resolve/resolve_dependency.m` — needs the depending package's FQN as context
- callers: `+mip/+dependency/find_all_dependencies.m`, `+mip/load.m`, `+mip/update.m`, `+mip/+state/check_broken_dependencies.m`
- tests in `tests/TestPackageDiscovery.m` (e.g. `testResolveDependency_BareNameIgnoresSameChannel` — current behavior would invert)

Related: #259 (install-time bare-name resolution via subscriptions), #263 (workflow for moving deps to core).

magland:
## Summary
- For bare-name deps, `mip.resolve.resolve_dependency` now takes the depending package's FQN as context. If the parent lives on `<owner>/<channel>` (non-core) and the dep is installed there, resolution returns that FQN; otherwise it falls back to `gh/mip-org/core/<name>`.
- Fully qualified deps still pass through unchanged.
- Callers updated to pass the parent FQN: `+mip/load.m`, `+mip/update.m`, `+mip/+dependency/find_all_dependencies.m`, `+mip/+state/check_broken_dependencies.m`.

This lets a package on `owner/chan` list bare deps for things that today live on `owner/chan` and tomorrow may move to `mip-org/core` — once the core version is what's installed, bare names automatically retarget there. Reviewers should manually load the core versions before approving the move (per the tradeoff acknowledged in the issue).

Out of scope (per the issue's touches list): the install-time path in `+mip/+dependency/build_dependency_graph.m` is tracked by #259.

Closes #264.

### DIFF
--- a/+mip/+dependency/find_all_dependencies.m
+++ b/+mip/+dependency/find_all_dependencies.m
@@ -2,8 +2,9 @@ function deps = find_all_dependencies(fqn, visited)
 %FIND_ALL_DEPENDENCIES   Recursively collect all transitive dependencies of an installed package.
 %
 % Reads mip.json from the installed package directory and resolves bare
-% dependency names to mip-org/core/<name>. Cycles are tolerated: a package
-% already in `visited` is not re-entered.
+% dependency names via mip.resolve.resolve_dependency, which prefers the
+% depending package's own channel before falling back to mip-org/core.
+% Cycles are tolerated: a package already in `visited` is not re-entered.
 %
 % Args:
 %   fqn     - Fully qualified package name (<owner>/<channel>/<name>)
@@ -44,7 +45,7 @@ try
     for i = 1:length(depNames)
         dep = depNames{i};
         try
-            depFqn = mip.resolve.resolve_dependency(dep);
+            depFqn = mip.resolve.resolve_dependency(dep, fqn);
         catch
             continue
         end
--- a/+mip/+resolve/resolve_dependency.m
+++ b/+mip/+resolve/resolve_dependency.m
@@ -1,15 +1,23 @@
-function depFqn = resolve_dependency(depName)
+function depFqn = resolve_dependency(depName, dependingPkgFqn)
 %RESOLVE_DEPENDENCY   Resolve a dependency name to a fully qualified name.
 %
 % If depName is already a FQN, return as-is.
-% If bare, resolve to mip-org/core/<name>.
 %
-% Bare-name dependencies always resolve to mip-org/core. To depend on a
-% package from a different channel, use the fully qualified name in
-% mip.yaml.
+% If depName is a bare name, prefer the depending package's own channel:
+% if the dep is installed at gh/<owner>/<channel>/<name> for the same
+% <owner>/<channel> as the depending package, return that FQN. Otherwise
+% fall back to gh/mip-org/core/<name>.
+%
+% This lets a package on owner/chan list bare deps that today live on
+% owner/chan and tomorrow may move to mip-org/core — once the core
+% version is what's installed, bare names automatically retarget there.
 %
 % Args:
-%   depName - Dependency name (bare or FQN)
+%   depName         - Dependency name (bare or FQN)
+%   dependingPkgFqn - (Optional) FQN of the package whose mip.yaml lists
+%                     this dep. Used only for bare-name resolution; if
+%                     omitted or not a gh FQN, bare names resolve
+%                     directly to mip-org/core.
 %
 % Returns:
 %   depFqn - Fully qualified name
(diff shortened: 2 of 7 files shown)


magland:
_Posting as Claude (Opus 4.7), automated PR review._

Overall this looks good — the new same-channel preference logic in [`+mip/+resolve/resolve_dependency.m`](https://github.com/mip-org/mip/blob/issue-264-prefer-own-channel-for-bare-deps/%2Bmip/%2Bresolve/resolve_dependency.m) is small, readable, and matches the spec in #264. All four call sites are updated consistently, and the tests cover the new branches (same-channel hit, fallback to core, core-parent skips the same-channel check, FQN passthrough still works). The optional second argument keeps backward compatibility for the existing one-arg test calls.

A couple of things worth noting (not blocking):

1. **Install/load divergence in [`+mip/+dependency/build_dependency_graph.m:65`](https://github.com/mip-org/mip/blob/issue-264-prefer-own-channel-for-bare-deps/%2Bmip/%2Bdependency/build_dependency_graph.m#L65).** The PR description correctly flags that the install path still hardcodes bare → `mip-org/core/<name>` and defers it to #259. The user-visible consequence is that for a package on `owner/chan` whose bare dep `B` is *only* available on `owner/chan` (not yet on core), `mip install` will try to install `mip-org/core/B` and fail (or install the wrong package), even though `mip load` afterward would have picked up `owner/chan/B`. That tradeoff is acknowledged in the issue, but the comment on line 16 of `build_dependency_graph.m` (`Bare-name dependencies are always resolved to mip-org/core/<name>.`) doesn't reference the new resolve_dependency behavior or #259. A one-line "see #259 — load-time path in `mip.resolve.resolve_dependency` already prefers same-channel" pointer there would save the next reader from being misled by the divergence.

2. **`find_all_dependencies` test coverage.** The behavioral change in `find_all_dependencies` (used by `mip update`, `mip unload`, and `prune_unused_packages`) is exercised indirectly via `testLoadPackage_BareDepPrefersOwnChannel`, but `TestFindAllDependencies.m` only uses `mip-org/core` packages, so it doesn't directly verify that a non-core parent's bare deps now traverse through same-channel installs. A small test there would lock in the change for the prune/update paths specifically.

Behaviors I checked and confirmed are correct:
- Non-`gh` parent FQNs (`local/`, `fex/`, etc.) correctly skip the same-channel branch and fall through to core (matches the docstring).
- `gh/mip-org/core/<parent>` correctly skips the same-channel branch via the explicit `~(owner=='mip-org' && channel=='core')` guard, including the dedicated `testResolveDependency_BareNameFromCoreParent` test.
- Recursion in `find_all_dependencies` correctly re-roots the parent FQN at each level (line 48 uses the *child's* fqn for its own dep resolution).
- `is_installed` canonicalizes the FQN, so name casing differences in `mip.yaml` don't break the same-channel hit.

No correctness blockers from me.

danfortunato:
Superseded — closing as redundant.

The bare-name own-channel resolution was independently implemented on `main` in 7754539 ("Resolve a package's bare-name deps within its own channel"), functionally equivalent to (and in places a superset of) this PR:

- `resolve_dependency(depName, parentFqn)` with the same "prefer the parent's non-core channel if the dep is installed there, else fall back to mip-org/core" logic. (main checks installation via `mip.resolve.installed_dir`; this branch used `mip.state.is_installed` — equivalent, since `is_installed` is itself built on `installed_dir` via `get_package_dir`.)
- The same caller wiring in `find_all_dependencies`, `check_broken_dependencies`, `load`, and `update`.
- Plus `docs/specification.md` §2.4.4–2.4.5 updates and `TestPackageDiscovery` unit tests (prefers-parent-channel, falls-back-when-absent, core-parent, no-parent) that this branch does not have.

The only delta is two `TestLoadPackage` integration tests; the resolver's unit-level coverage on `main` is considered sufficient. Closing #264 as completed.

PR #281, opened 2026-06-17
Install-time x86-64 SIMD variant selection

magland:
Detects the host CPU's x86-64 microarchitecture level and installs the highest published build variant it supports.

- `+mip/+build/detect_cpu_level.m`: pure-MATLAB level detection via `/proc/cpuinfo` (Linux) and a bundled `cpu_level.ps1` using kernel32 `IsProcessorFeaturePresent` (Windows). No MEX, so it runs before any compiled package is installed.
- `+mip/+build/compatible_archs.m`: ordered architecture preference list — highest supported SIMD level first, then the base arch, then `any` (the numbl_wasm fallback is preserved).
- `select_best_variant` now walks that preference list; `build_package_info_map` and `avail` use it.
- `mip.build.arch()` still returns the base arch, so local builds and `match_build` are unaffected and a pre-SIMD client simply ignores unknown `_vN` index entries (verified).
- Tests for detection, the preference list, and SIMD-aware selection (the Windows `.ps1` is exercised on the windows-latest CI leg).

Backward compatible — this can land independently. The SIMD builds it selects are produced by the mip-core and mip_channel_tools PRs.

---

Part of one change, in merge order:
1. mip-org/mip_channel_tools#1 — build-system SIMD support
2. mip-org/mip-core#33 — finufft SIMD build variants
3. mip-org/mip#281 — client-side variant selection (this PR)


magland:
@danfortunato 
@DiamonDinoia

DiamonDinoia:
the linux follows the standard psABI with 4 levels but on windows only 3 are used. Is there any reason for that?

magland:
(Claude here, replying on behalf of @magland.)

It's an MSVC limitation, not a deliberate choice. On Windows we compile with MSVC, whose `/arch:` granularity is coarser than GCC's `-march`: for x64 it only accepts `AVX`, `AVX2`, and `AVX512` — there's no `/arch:` for the v2 feature set (SSE4.2/POPCNT/etc.), since SSE2 is the implicit x64 baseline (`/arch:SSE2` is x86-only). The runtime detector (`IsProcessorFeaturePresent`) likewise has no reliable SSE4.2 id.

So a "v2" Windows build would be byte-for-byte the SSE2 baseline. Windows therefore ships baseline (≈v1) + v3 (`/arch:AVX2`) + v4 (`/arch:AVX512`), and a v2-only CPU correctly falls back to the baseline build. Linux/GCC can target all four psABI levels distinctly, so it does.

DiamonDinoia:
I see, sounds good to me then. I can't speak for the matlab but the flags
chosen look correct to me.

PR #284, opened 2026-06-22
Auto-bundle MEX runtime libraries after compile

magland:
`prepare_package` now vendors each compiled MEX file's runtime-library dependencies (e.g. `libgfortran`, `libgomp`) automatically after the compile script runs, via the new `mip.build.bundle_mex_libs`. Previously each package's `compile.m` had to discover its MEX files and call the bundler itself.

The platform-specific work is delegated to `bundle_runtime_libs`, which the channel build environment puts on the MATLAB path (`mip_channel_tools/scripts`). When that function is absent — e.g. a standalone `mip bundle` outside a channel build — `bundle_mex_libs` is a no-op, so the package manager keeps no hard dependency on the channel tooling. This is the post-compile counterpart to the existing pre-compile `strip_prebuilt_binaries` step.

Channel `compile.m` scripts that currently call `bundle_runtime_libs` directly can drop those calls (separate channel-repo changes).

### DIFF
--- a/+mip/+build/bundle_mex_libs.m
+++ b/+mip/+build/bundle_mex_libs.m
@@ -1,0 +1,28 @@
+function bundle_mex_libs(pkgDir)
+%BUNDLE_MEX_LIBS   Vendor runtime-library dependencies for compiled MEX files.
+%
+% After a package's compile script runs, each MEX file it produced may depend
+% on non-system shared libraries (e.g. libgfortran, libgomp) that are not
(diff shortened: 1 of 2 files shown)


magland:
Companion PR: mip-org/mip_channel_tools#3 makes `copy_and_sanitize_lib` idempotent so re-bundling the same MEX is safe. Land that one (and update channel `compile.m` scripts) alongside this — the build workflow checks out `mip-org/mip` at `main`, so this needs to be in before channels drop their explicit `bundle_runtime_libs` calls.

PR #289, opened 2026-06-25
Skip MATLAB BLAS/LAPACK libs when bundling MEX runtime deps

magland:
Extract the `bundle_runtime_libs` skip-lists into a shared `mip.build.runtime_skip_libs` helper and add `libmwblas`/`libmwlapack` to both the Linux SONAME list and the macOS install-path prefixes.

MATLAB ships and resolves these at runtime (like `libmx`/`libmex`), so they must never be vendored. On macOS their `LC_LOAD_DYLIB` is `@rpath/libmwblas.dylib`, which has no on-disk path, so the bundler would hard-fail trying to copy them — this broke any package that links MATLAB's BLAS (e.g. the new `fmm3dbie` package). Adds `tests/TestRuntimeSkipLibs.m`.

### DIFF
--- a/+mip/+build/bundle_runtime_libs.m
+++ b/+mip/+build/bundle_runtime_libs.m
@@ -60,7 +60,7 @@ for i = 1:numel(m)
     resolved(m{i}{1}) = m{i}{2};
 end
 
-skip = linux_skip_set();
+skip = mip.build.runtime_skip_libs();
 bundled = false;
 for i = 1:numel(needed)
     so = needed{i};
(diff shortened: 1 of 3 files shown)


magland:
Superseded by #285 (transitive-runtime-libs): bundle_runtime_libs now detects MATLAB-provided libraries dynamically via scan_matlab_libnames(matlabroot/bin/<arch>), so libmwblas/libmwlapack are already skipped on both Linux and macOS. No static skip-list change needed.

PR #309, opened 2026-07-01
Fix: mip update --no-compile mip must error instead of self-updating

### LINKED ISSUE #303: mip update --no-compile mip performs a destructive self-update instead of erroring
Spec §7.6 says a batch containing `mip` itself must raise `mip:update:noCompileRequiresEditable` **before any destructive action**. It doesn't.

The `--no-compile` guard only checks `kind=='process'` items ([update.m:126](https://github.com/mip-org/mip/blob/main/+mip/update.m#L126)), but the mip identity is classified `self-update` ([update.m:219](https://github.com/mip-org/mip/blob/main/+mip/update.m#L219)) and falls through to the `updateSelf` hot-swap ([update.m:156](https://github.com/mip-org/mip/blob/main/+mip/update.m#L156)), which takes no `noCompile` argument.

Result: unpinned `mip update --no-compile mip` silently reinstalls mip instead of erroring. `TestUpdateSelf` only covers `--force mip`, so this is untested.

Fix: reject the mip identity (and any non-editable-local item) in the `--no-compile` pre-pass before the walk.

magland:
The `--no-compile` guard in `+mip/update.m` only checked `kind=='process'` items, so `mip update --no-compile mip` (classified `kind=='self-update'`) skipped validation and fell through to the destructive `updateSelf` hot-swap.

The guard now rejects every actionable item (both `process` and `self-update`) that is not an editable local install, raising `mip:update:noCompileRequiresEditable` before any destructive action per spec §7.6, while still leaving `pin-skip`/`no-source-skip` items untouched.

Adds a `TestUpdateSelf` case asserting both `mip update --no-compile mip` and the `--force` variant raise the error before any network or destructive action.

Closes #303

### DIFF
--- a/+mip/update.m
+++ b/+mip/update.m
@@ -118,16 +118,17 @@ function update(varargin)
         items{i} = classifyArg(args{i});
     end
 
-    % --no-compile only applies to editable local installs. Validate
-    % across all actionable items before any destructive action. The mip
-    % self-update (kind 'self-update') is never an editable local install,
-    % so it must be rejected here too rather than silently hot-swapping mip.
+    % --no-compile only applies to editable local installs. Validate every
+    % actionable item before any destructive action; the mip self-update is
+    % never an editable local install, so it is rejected here rather than
+    % falling through to the destructive updateSelf hot-swap. Restricting to
+    % the actionable kinds matters: the skip kinds carry no .pkg (pin-skip)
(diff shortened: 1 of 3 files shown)


magland:
@danfortunato

I'm having trouble wrapping my head around this one from claude. Do you understand this?

PR #310, opened 2026-07-01
docs: reconcile download integrity/security (SHA-256 suspended, HTTPS-only)

### LINKED ISSUE #307: docs: reconcile download integrity/security (SHA-256 verification, HTTPS-only)
Two download-integrity/security behaviors are out of sync with the spec.

**SHA-256 verification (advertised but disabled)**
- The verification block is commented out (per #201) in [download_mhl.m:87-108](https://github.com/mip-org/mip/blob/main/+mip/+channel/download_mhl.m#L87-L108), and `tests/TestDownloadMhl.m` asserts a *wrong* digest does **not** error.
- But §3.6, §3.1.6, and §13 (`mip:digestMismatch`) present it as an active guarantee. Either re-enable verification or mark it suspended in the spec so it doesn't advertise a protection that isn't there.

**HTTPS-only (enforced but undocumented)**
- Plain `http://` is rejected: `mip:downloadMhl:requireHttps` ([download_mhl.m:39](https://github.com/mip-org/mip/blob/main/+mip/+channel/download_mhl.m#L39)) and `mip:install:requireHttps` for `--url` ([install.m:922](https://github.com/mip-org/mip/blob/main/+mip/install.m#L922)).
- Not mentioned in §3.3/§3.4 prose or the §13 error table.

magland:
Reconciles `docs/specification.md` with the actual download integrity/security behavior in code.

**SHA-256 digest verification is currently suspended.** `+mip/+channel/download_mhl.m` has the verification block commented out (see #201): channel publishing is not atomic — the `.mhl` asset is uploaded before `index.json` is redeployed, so a client fetching mid-publish would pair an old digest with a new archive and wrongly trip `mip:digestMismatch`. The spec previously described verification as active. Updated §3.1.6, §3.6, and the §13 `mip:digestMismatch` row to state it is not currently performed (index may still carry `mhl_sha256` for when it is re-enabled). The active path-traversal validation in §3.6 is left unchanged.

**HTTPS-only downloads are enforced but were undocumented.** Plain `http://` is refused for remote `.mhl` sources (`mip:downloadMhl:requireHttps`) and for `--url` zip installs (`mip:install:requireHttps`). Added notes to §3.3 and §3.4 and both error IDs to the §13 table.

Closes #307

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -312,7 +312,7 @@ Priority: exact match > `numbl_wasm` fallback > `any`.
 1. For each package in topological order:
    - If already installed (directory exists), skip it.
    - Download the `.mhl` file from the URL in the index.
-   - If the channel index entry includes `mhl_sha256`, verify the downloaded file against it. Mismatch raises `mip:digestMismatch` and deletes the corrupted file. Missing digest (legacy releases) or an unavailable JVM (e.g. `numbl_wasm`) silently skips verification. See [§3.6](#36-mhl-archive-integrity).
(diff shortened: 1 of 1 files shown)


magland:
@danfortunato this looks good to me

PR #313, opened 2026-07-01
docs: correct parsing, naming, and dependency-resolution rules

### LINKED ISSUE #306: docs: §1-§2 parsing, naming, and dependency-resolution rules contradict the implementation
Several §1–§2 rules contradict the implementation and tests (spec is stale in each case).

**Bare-name dependency resolution**
- §3.1.5 and §4.4 say bare deps "always resolve to `gh/mip-org/core`" — §4.4 even cites §2.4.4 while stating its opposite. Code resolves to the parent's own channel for non-core parents ([resolve_dependency.m:40-52](https://github.com/mip-org/mip/blob/main/+mip/+resolve/resolve_dependency.m#L40-L52), [build_dependency_graph.m:83-98](https://github.com/mip-org/mip/blob/main/+mip/+dependency/build_dependency_graph.m#L83-L98)), matching §2.4.4/§2.4.5 and `tests/TestPackageDiscovery.m`. Fix the §3.1.5/§4.4 wording.

**2-part personal-channel shorthand**
- §2.3 says `<owner>/<name>` is `--channel`-only and "does not apply to FQN package arguments"; §2.1 maps a 2-part arg to a non-gh FQN. But `foo/pkg` → `gh/foo/foo/pkg` ([parse_package_arg.m:99-114](https://github.com/mip-org/mip/blob/main/+mip/+parse/parse_package_arg.m#L99-L114)); `tests/TestUtilsParsing.m` requires it.
- Display form collapses `gh/<o>/<o>/<pkg>` → `<o>/<pkg>` ([display_fqn.m:29-33](https://github.com/mip-org/mip/blob/main/+mip/+parse/display_fqn.m#L29-L33)), which §1.1.1 ("strips `gh/` only") omits.

**Name validation**
- §1.1/§2.1 give `^[-a-zA-Z0-9_.]+$` and say `.github` is valid, but [is_valid.m:28](https://github.com/mip-org/mip/blob/main/+mip/+name/is_valid.m#L28) forbids dots and leading/trailing separators; `tests/TestNameValidation.m` asserts `.github`/`my.pkg` are invalid.
- §9.8 says `mip init --name` allows uppercase and isn't lowercased, but [is_valid_canonical.m:29](https://github.com/mip-org/mip/blob/main/+mip/+name/is_valid_canonical.m#L29) is lowercase-only, so `mip init --name MyPkg` errors (`tests/TestInit.m:147`).

magland:
Corrects several places in `docs/specification.md` where the spec disagreed with the implemented behavior (verified against the code and the unit tests).

- **Bare-name dependency resolution** (§3.1.5, §4.4, and the §3.1.2 rationale): these said bare deps "always resolve to `gh/mip-org/core`", directly contradicting §2.4.4/§2.4.5 (and §4.4 even cited §2.4.4 while stating its opposite). Updated to the actual parent-channel-preference rule: a bare dep resolves to the depending package's own channel when the parent is a `gh` package on a non-core channel that provides the dep, otherwise to `gh/mip-org/core/<name>`. Matches `resolve_dependency.m` / `build_dependency_graph.m` and `TestPackageDiscovery`.
- **Personal-channel shorthand** (§2.1, §1.1.2, §2.3): a 2-part `<owner>/<name>` package argument (first segment not a reserved type) parses as `gh/<owner>/<owner>/<name>`. Added the table rows and removed §2.3's false claim that this shorthand does not apply to package arguments. Matches `parse_package_arg.m` / `TestUtilsParsing`.
- **Display form** (§1.1.1): documented the `gh/<owner>/<owner>/<pkg>` -> `<owner>/<pkg>` collapse (skipped when the owner is a reserved type), per `display_fqn.m`.
- **User-input name validation** (§1.1, §2.1): replaced `^[-a-zA-Z0-9_.]+$` and the "`.github` is valid" example with the real `is_valid` rule (`^[a-zA-Z0-9]([-a-zA-Z0-9_]*[a-zA-Z0-9])?$` — no dots; must start/end alphanumeric). Matches `TestNameValidation`.
- **`mip init --name`** (§9.8): corrected the regex to the canonical lowercase rule `^[a-z0-9]([-a-z0-9_]*[a-z0-9])?$` and fixed the claim about `--name` not being lowercased — an uppercase `--name` (e.g. `MyPkg`) raises `mip:init:invalidName`. Matches `is_valid_canonical.m` / `init.m` / `TestInit`.

Closes #306

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -21,15 +21,17 @@ For GitHub channel packages:
 - **channel** -- the channel name within that owner
 - **name** -- the package name
 
-All components must match the regex `^[-a-zA-Z0-9_.]+$` and must not be `.` or `..`.
+All components must match the regex `^[a-zA-Z0-9]([-a-zA-Z0-9_]*[a-zA-Z0-9])?$` — letters, digits, hyphens, and underscores, starting and ending with a letter or digit. Dots are not allowed anywhere, so `.`, `..`, and dotted names like `.github` are all rejected. See [§2.1](#21-parsing-a-package-argument-parse_package_arg) for the full validation rules.
 
 #### 1.1.1 Display Form
 
-The **display form** of a canonical FQN strips the `gh/` prefix only; non-gh FQNs are displayed unchanged. `mip list`, `mip info`, and all install/load/update messages use the display form. `mip.parse.display_fqn` performs the conversion.
+The **display form** of a canonical FQN strips the `gh/` prefix; non-gh FQNs are displayed unchanged. For **personal channels** — where the channel equals the owner (`gh/<owner>/<owner>/<pkg>`) — the duplicated owner segment is additionally collapsed, yielding the 2-part form `<owner>/<pkg>`. The collapse is skipped when the owner matches a reserved source-type prefix (`gh`, `local`, `fex`, `web`, `mhl`), since the collapsed form would otherwise be re-parsed as a non-gh FQN rather than round-tripping to the original. `mip list`, `mip info`, and all install/load/update messages use the display form. `mip.parse.display_fqn` performs the conversion.
 
 | Canonical FQN | Display form |
 |---|---|
 | `gh/mip-org/core/chebfun` | `mip-org/core/chebfun` |
+| `gh/magland/magland/chunkie` | `magland/chunkie` (personal-channel collapse) |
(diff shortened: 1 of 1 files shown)


magland:
@danfortunato this looks good to me

PR #332, opened 2026-07-02
Enforce mip.yaml version semantics

### LINKED ISSUE #315: mip version semantics
Here is what Jeremy and I decided.

The version field in mip.yaml is optional. It can either be blank, or not appear at all. If the version field is non-empty, then the version string must be numeric. If the version string is non-numeric, we need to throw an error somewhere (e.g. when bundling).

When published to a channel, the following rules apply:
- If the mip.yaml contains a numeric version, then the subfolder name specifying the version in the channel must either match that numeric version, or be non-numeric.
- If the channel subfolder version is non-numeric, then that takes precedence over the yaml version string. This is the case where the channel publishes a "main" version, but the upstream mip.yaml might contain a numeric version string. Main takes precedence.

The version string is written to mip.json, which is always the canonical place that mip reads the version from. mip never reads the version from mip.yaml for MHL installations.

For editable installs, mip may read the version from mip.yaml.

magland:
Implements the version semantics decided in #315:

- `mip.yaml`'s `version` is optional; when non-empty it must be numeric. `read_mip_yaml` now rejects non-numeric values and normalizes unquoted numeric YAML scalars to strings, so `mip.json` always records a string version.
- `prepare_package` enforces the channel rules for the `.release_version` side file: a non-numeric release version (e.g. `main`) takes precedence over the yaml version; a numeric one must match it (`mip:build:versionMismatch` otherwise).
- `mip.json` remains the canonical version source and now always carries a non-empty string: a blank version with no channel override is recorded as `unspecified` (bundle filenames become `pkg-unspecified-any.mhl` instead of `pkg--any.mhl`).
- Editable installs read the version live from the source `mip.yaml` in `read_package_json`, falling back to the `mip.json` snapshot when the source yaml is missing or invalid.
- `is_numeric_version` tightened to digit-only components (previously `1e3`, `-1`, `inf` passed via `str2double`), matching the channel build's definition.
- Spec updated (§1.3, §11.1, §11.2.1).

Companion PR in mip_channel_tools aligns `validate_channel_version_rules` with the same rules.

Closes #315


danfortunato:
>  Editable installs read the version live from the source mip.yaml in read_package_json, falling back to the mip.json snapshot when the source yaml is missing or invalid.

Do editable installs actually have a mip.json to fall back to?

danfortunato:
`mip init` currently initializes mip.yaml with `version: ""`. is that the default we want for `mip init`? another option is to default to not including the version line at all

magland:
> `mip init` currently initializes mip.yaml with `version: ""`. is that the default we want for `mip init`? another option is to default to not including the version line at all

I'm okay either way. The advantage of version: "" is that user knows they can put in a version if they want. But maybe version: with no quotes would be better.

I don't have a strong opinion.

danfortunato:
let's leave it as is then

but now I'm forgetting why the version cannot be non-numeric... what is the problem if mip.yaml has `version: "main"` again?

PR #335, opened 2026-07-02
Gate mip-core dispatch on the test matrix passing

danfortunato:
Follow-up to #334, which merged the initial version before this gating change was added. Right now `main` has a standalone `dispatch-core-build.yml` that fires on **every** push to `main` without waiting for tests. This PR replaces it with a gated job.

## What changes
- Removes the standalone `dispatch-core-build.yml`.
- Adds a `dispatch-core-build` job **to `tests.yml`** that dispatches `build-package.yml` on `mip-org/mip-core` for `packages/mip/main` (arch `any`, `force=false`).

## Gating
- **`needs: test`** — runs only if the full multi-OS test matrix passes. The matrix is `fail-fast: false`, so every arch runs; a failure on **any** arch fails `test` and skips the dispatch.
- **Not `needs: coverage`** — a codecov upload or low-coverage failure must **not** block a release, so the dispatch ignores the coverage job.
- **`if: push to main`** — pull requests (including forks) run tests but never run this job, so the `repo`-scoped `MIP_CORE_DISPATCH_TOKEN` is never exposed to PR code.

## Verify after merge
Push a trivial change to `main`; once the test matrix is green, confirm a new **Build Package** run appears in `mip-org/mip-core` for `packages/mip/main`. A red arch should produce no dispatch.


magland:
LGTM

Issue #345, opened 2026-07-16
Re-loading an already-loaded package refreshes recency but not path precedence (last-loaded doesn't win shadowing)


danfortunato:
## Summary

Re-loading an already-loaded package (`mip load A` when `A` is already loaded) refreshes its **recency** in the state lists but leaves the **physical MATLAB path unchanged**. As a result, "most recently loaded wins" holds for every list-based consumer (bare-name resolution, ambiguous unload, `mip list` sort) but silently *does not* hold for the one place the phrase is most literal — function shadowing. Re-loading `A` to make its functions win doesn't actually promote them.

## Steps to reproduce

Two installed packages `A` and `B` that both ship a `foo.m` (they shadow each other):

```matlab
mip load A        % path: [A];      A/foo wins
mip load B        % path: [B, A];   B/foo wins  (addpath prepends -> B in front)
mip load A        % re-load: A is now "most recently loaded"
which foo         % -> still B/foo   <-- surprising
```

**Expected:** after `mip load A`, `A` is the most recently loaded package, so `A/foo` should win.
**Actual:** `B/foo` still wins — the re-load never touched the path.

The only way to actually make `A` win today is `mip unload A && mip load A`, which removes A's entries and re-prepends them.

## Root cause

Two orderings are maintained, and they are reverses of each other:

- **Recency lists** (`MIP_LOADED_PACKAGES` / `MIP_DIRECTLY_LOADED_PACKAGES`) — newest is **appended** (end of list). Drives bare-name resolution, ambiguous unload, `mip list`.
- **Physical path** (`addpath`, which **prepends**) — newest is at the **front**. Drives function shadowing.

With no re-loads, these stay consistent: `path == reverse(recency list)`. But the already-loaded branch in `+mip/load.m` bumps the package to the end of the lists and then **returns early, before the `addpath` step** — so a re-load moves the entry in the list but not on the path, and the two orderings desync. The recency-bump behavior was added for the list consumers (see the `#267` recency work); path precedence was simply never in scope.

This is internally consistent with the spec as written (§4.1 step 5 and the tiebreaker sections only define "most recent" for the state lists), so it is not a spec *violation* — but it is a real inconsistency in the "last loaded wins" mental model.

## Proposed fix

[...]

PR #346, opened 2026-07-20
Implement MEP 8: environments (mip env, mip activate/deactivate)

### LINKED ISSUE #161: Local package updates should install missing dependencies instead of erroring
## Problem

When updating a local package via `mip update`, the update calls `mip.build.install_local` which validates that all dependencies are installed and errors if any are missing (`mip:dependencyNotFound` at `+mip/+build/install_local.m:51`).

Remote package updates handle this differently: after downloading the new version, `installMissingDeps` runs and installs any dependencies that aren't present.

This means:
- `mip update mip-org/test-channel1/gamma` (remote) with a missing dep → dep is automatically installed
- `mip update local/local/mypkg` (local) with a missing dep → errors out

## Expected behavior

Both paths should be consistent. Local package updates should install missing dependencies from the channel rather than erroring.

## Context

Found during review of #137. The `installMissingDeps` step already exists for remote updates — the local update path just doesn't use it because `install_local` validates deps before it gets that far.

magland:
Implements [MEP 8](https://github.com/mip-org/meps/blob/main/meps/mep-0008.md).

- `mip env create/list/delete` for named envs (`<baseline root>/envs/<name>`) and path envs (`./.mip`); an env is any directory with a `packages/` subtree
- `mip activate` / `mip deactivate` (aliases for `mip env activate/deactivate`) move the session pointer via `MIP_ROOT` with a full load-state swap and restore; `--load` loads the env's directly installed packages
- Session commands act on the active env and print a leading `environment:` line while one is active
- Self-uninstall/self-update trigger only when the active root is the root mip runs from; elsewhere `gh/mip-org/core/mip` is an ordinary, inert package
- Bulk unloads never unload the package providing the running mip, so a loaded preview build survives activation
- Local installs auto-install missing channel dependencies (closes #161)

Spec sections 5/14, tests, and changelog updated. Preview published as `mip-org/labs/mip@mep-0008-b`.


danfortunato:
@magland this issue came up when I was going through MEP 8 last week: #345

I think this is important for MEP 8, since we need to restore the path state to what it was after an env is deactivated

magland:
> @magland this issue came up when I was going through MEP 8 last week: #345
> 
> I think this is important for MEP 8, since we need to restore the path state to what it was after an env is deactivated

Yeah you're right. Do you want me to take a crack at incorporating this into the document and the PR?

danfortunato:
> > @magland this issue came up when I was going through MEP 8 last week: #345
> > I think this is important for MEP 8, since we need to restore the path state to what it was after an env is deactivated
> 
> Yeah you're right. Do you want me to take a crack at incorporating this into the document and the PR?

Yes!

magland:
#349 fixes #345 (re-loading a package re-applies its paths, so `path == reverse(load order)` always holds). `mip deactivate` relies on that to restore the pre-activation path exactly.

Assuming #349 merges, I've updated this branch's spec (§4.1 step 5, §14.4) and the MEP doc (mip-org/meps#16, merged) to document that dependency.

+mip/+env/baseline_root.m
```diff
@@ -0,0 +1,35 @@
+function root = baseline_root()
```
danfortunato: I wonder if "baseline" should be renamed to "base"

+mip/+env/is_path_arg.m
```diff
@@ -0,0 +1,17 @@
+function tf = is_path_arg(arg)
+%IS_PATH_ARG   Syntactic name-vs-path disambiguation for env arguments.
+%
+% Usage:
+%   tf = mip.env.is_path_arg(arg)
+%
+% A bare word is an environment name; anything containing a path
+% separator ('/', or '\' on Windows) is a path. The two are never
```
danfortunato: We could mirror `mip.install.is_local_path` and require a path to an env to begin with `~`, `.`, `/`, or a drive letter for consistency. Then they can share a utility.

danfortunato: We decided that specifying env paths can be more general than specifying package install paths (e.g. `mip activate foo/bar` is allowed). However, currently `is_path_arg` returns false for the string `'.mip'`.

magland:
This looks good to me. There were some potential edge case issues in claude's review, but those will be fixed when we disallow mip itself to be installed in an environment.

I tested it via the newly built package on labs

mip install mip-org/labs/mip@mep-0008-b

And things seems to be working for both types of environments.

PR #354, opened 2026-07-27
Enforce the self-scenarios rules for mip as a package

magland:
Implements `docs/mip_self_scenarios.md` in the spec and the implementation.

- New `mip.self.op_state` classifies the session (`ok` / `standalone` / `env` / `mip-loaded`). The self operations — `mip update mip`, `mip uninstall mip`, `mip install mip@<version>` — proceed only in `ok`: standalone sessions get a message refusal, an active environment raises `mip:self:envActive`, a loaded other mip raises `mip:self:otherMipLoaded`.
- The loaded package providing the running mip can no longer be updated or uninstalled (`mip:update:runningMip` / `mip:uninstall:runningMip`); `mip update --all` skips guarded packages instead of erroring.
- No mip package — core or otherwise — may be installed into or loaded from an environment (`mip:env:noMip`), across all install types and loads.
- Activation keeps a running preview mip's code on the path but drops it from the loaded-package lists; deactivation restores it as a loaded package with its prior flags.
- `mip load mip` is the always-loaded no-op even in roots holding no copy.
- The self-update notice prints only in the `ok` state, so it never suggests a command mip would refuse.

Spec: new §1.7.1 (self-operation preconditions) plus updates to §3, §4.1, §5.4.1, §6, §7, §11.8, §13, §14.

Tests: `TestSelfOpGuards` and `TestEnvNoMip` are new; `TestRunningMipGuard`, `TestEnvSelfGuard`, and `TestMipUpdateNotice` updated. `TestMipUpdateNotice` is now registered in `run_tests.m` — it was previously never run.

One deliberate divergence: bare `mip uninstall mip` with another mip installed still hits the §6.3 ambiguity refusal (listing FQNs) rather than `mip:self:otherMipLoaded`; the FQN form gets the guard error. Both refuse safely.


danfortunato:
Unlanded from `main`: this was merged before #352 (the revision of `docs/mip_self_scenarios.md` that this PR implements against) had landed, and the implementation does not match the revised spec on Scenario 10.

`main` was force-pushed back to `6819cde` (the #346 merge). The merge commit `26bc315` is preserved at the tag `rollback/pr-354`.

GitHub does not allow reopening a merged PR, so work continues in #355 — same head commit, unchanged diff, with the outstanding deltas versus #352 written up in the description.

Issue #351, opened 2026-07-23
Require FQN when a package is installed from multiple channels


magland:
If the same package is installed from more than one channel, bare-name resolution for `mip load` and `mip update` should refuse and ask for the FQN, instead of the current §2.4.1 priority rule (mip-org/core first, then alphabetical). Spec §2.4.1 needs to change accordingly.

`mip unload <name>` is the exception: it should unload the most recently loaded package of that name (current §2.4.2 behavior). `mip uninstall` already refuses ambiguous bare names (§2.4.3).

PR #355, opened 2026-07-29
Enforce the self-scenarios rules for mip as a package

danfortunato:
Reopens #354, which was merged prematurely and has been unlanded from `main`.

GitHub does not allow reopening a merged PR, so this is a new PR from the same
head commit (`26f60ab`) — the diff is unchanged. `main` was force-pushed back to
`6819cde` (the #346 merge); the original merge commit is preserved at the tag
`rollback/pr-354` if it is ever needed.

**Why it was unlanded:** #354 was implemented against the version of
`docs/mip_self_scenarios.md` that landed with #346, but #352 revises that
document and was still open. #352 should land first; this PR then needs the
changes below to match it.

## Outstanding work to match #352

**1. Scenario 10 — `mip unload --all` (contradiction).** #352 says the bulk
unload must also unload a loaded non-core mip, with only the main mip spared,
which then becomes the active mip again. This PR implements the opposite:
`unloadAll` filters through `isRunningMip` (`+mip/unload.m`), `docs/specification.md`
§5.4.1 codifies "the running mip is never bulk unloaded", and
`tests/TestRunningMipGuard.m` has two tests (`testUnloadAllForceSparesPreview`,
`testUnloadAllSparesNonStickyPreview`) asserting the old rule.

Two things to be careful about when changing this:

- The §5.4.1 exemption is shared with `mip reset`, the activate/deactivate swap
  and dependency pruning. #352 only changes `mip unload --all`, so the exemption
  needs to be split, not removed.
- Unloading the running mip pulls the executing code off the path mid-command.
  The explicit-unload path already handles this by capturing `runningMip` before
  any `rmpath`; `unloadAll` will need the same treatment.

**2. Scenarios 8 and 11 — bare-name ambiguity (not implemented).** #352 says
that with mip installed from two channels, bare `mip load mip` and bare
`mip update mip` are refused as ambiguous and require the FQN. Today:

- `mip load mip` → `resolve_bare_name` applies the §2.4.1 priority rule, picks
  the core identity, and prints "always loaded" — a silent no-op, not a refusal.
- `mip update mip` → the same resolution makes it a self operation, so it raises
  `mip:self:otherMipLoaded`. It refuses, but for the wrong reason.

This is blocked on #351, which asks for exactly this change to §2.4.1. Either
land #351 first or implement the refusal as part of that issue.

**3. Minor deviations.** Scenario 12's bare-name ambiguity refusal is a printed
message plus `continue` rather than a thrown error (pre-existing §6.3 behavior).
Scenario 22's bare `mip update mip` under an active environment raises
`mip:self:envActive` instead of resolving against the environment's root as #352's
wording implies; arguably the more helpful message, but worth a decision.

## What already matches #352

The state-based definitions of main mip / main root (nothing in the
self-detection path consults the saved path); Scenario 9 (bare `mip unload mip`
unloads the most recently loaded mip); the FQN forms of Scenarios 11 and 12,
including both error identities and the "unload `other/channel/mip` first"
guidance; Scenario 12's bare-name ambiguity listing; and Scenario 22's FQN forms
for update, uninstall and `install@<version>`.

Note one correction to #354's original description: it listed bare
`mip uninstall mip` hitting the §6.3 ambiguity refusal as "one deliberate
divergence". That was a divergence from the pre-#352 doc — under #352 it is the
specified behavior, so it is a match.

## Original description

Implements `docs/mip_self_scenarios.md` in the spec and the implementation.

- New `mip.self.op_state` classifies the session (`ok` / `standalone` / `env` / `mip-loaded`). The self operations — `mip update mip`, `mip uninstall mip`, `mip install mip@<version>` — proceed only in `ok`: standalone sessions get a message refusal, an active environment raises `mip:self:envActive`, a loaded other mip raises `mip:self:otherMipLoaded`.
- The loaded package providing the running mip can no longer be updated or uninstalled (`mip:update:runningMip` / `mip:uninstall:runningMip`); `mip update --all` skips guarded packages instead of erroring.
- No mip package — core or otherwise — may be installed into or loaded from an environment (`mip:env:noMip`), across all install types and loads.
- Activation keeps a running preview mip's code on the path but drops it from the loaded-package lists; deactivation restores it as a loaded package with its prior flags.
- `mip load mip` is the always-loaded no-op even in roots holding no copy.
- The self-update notice prints only in the `ok` state, so it never suggests a command mip would refuse.

Spec: new §1.7.1 (self-operation preconditions) plus updates to §3, §4.1, §5.4.1, §6, §7, §11.8, §13, §14.

Tests: `TestSelfOpGuards` and `TestEnvNoMip` are new; `TestRunningMipGuard`, `TestEnvSelfGuard`, and `TestMipUpdateNotice` updated. `TestMipUpdateNotice` is now registered in `run_tests.m` — it was previously never run.


danfortunato:
#352 has landed on `main` (`df00674`), so `docs/mip_self_scenarios.md` is now the revised spec. This PR still merges cleanly onto it (no overlap — it never touches that file), and the three outstanding items in the description are unchanged: Scenario 10 is the contradiction to fix, Scenarios 8/11 are blocked on #351, and the two minor deviations need a decision.

